PR #225, opened 2013-08-23
add auto reconnect in execute_sql

beanyoung:
I use peewee with tornado, and all requests share one database connection, but when something wrong happens with database, the connection will be unavaliable. So I added a `try exception` block in `execute_sql()`, and this works fine for me.


coleifer:
In case you're interested, I've got a mixin for retrying failed queries in 017e4e4952f0b429dd21f0e90465a0f644cf6015 . Docs:

http://docs.peewee-orm.com/en/latest/peewee/database.html#automatic-reconnect

PR #363, opened 2014-05-28
Created a new parameter 'reconnect' to indicate if MySQLAdapter can try ...

andmart:
Created a new parameter 'reconnect' to indicate if MySQLAdapter can try reconnect to server. 
In long time connections, timeout makes connection to drop.
I have an app that faces too many "MySQL server has gone away" errors. So, this can try reconnect to the server in case of timeout.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -279,7 +279,7 @@ class Database(object):
             return func(self, *args, **kwargs)
         return inner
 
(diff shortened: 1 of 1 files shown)


coleifer:
In case you're interested, I've got a mixin for retrying failed queries in 017e4e4952f0b429dd21f0e90465a0f644cf6015 . Docs:

http://docs.peewee-orm.com/en/latest/peewee/database.html#automatic-reconnect

PR #494, opened 2015-01-05
[fix] TypeError: host is an invalid keyword argument for this function

ekatsah:
Try to fix this:

    self = <peewee.SqliteDatabase object at 0x7f79bbfcf950>, database = ':memory:'
    kwargs = {'host': file.db'}

    def _connect(self, database, **kwargs):
    >       conn = sqlite3.connect(database, **kwargs)
    E TypeError: 'host' is an invalid keyword argument for this function

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -56,4 +56,10 @@ def connect(url):
     if database_class is MySQLDatabase and 'password' in connect_kwargs:
         connect_kwargs['passwd'] = connect_kwargs.pop('password')
 
+    # Adjust parameters for SQLite
+    if (database_class is SqliteDatabase and 'host' in connect_kwargs and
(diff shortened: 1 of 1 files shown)


coleifer:
I'm confused -- why is `host` getting used?

ekatsah:
Hum i read the doc too rapidly and confused relatif/absolute path. please discard this PR. sorry for the noise.

Was trying to fix this kind of error

    self = <peewee.SqliteDatabase object at 0x7fc15db82c90>
    database = 'tmp/file.db', kwargs = {}

    def _connect(self, database, **kwargs):
    >       conn = sqlite3.connect(database, **kwargs)
    E OperationalError: unable to open database file

but the doc is right. furthermore, this fix doesnt cover the use case (absolute path...)

PR #505, opened 2015-01-22
playhouse.migration bug fix

d1hotpep:
this variable was mis-labeled such that migrations crashed if any column contained extra attributes, like 'auto_increment'.  this fixes it


coleifer:
Thanks!

PR #508, opened 2015-01-22
primary null values

d1hotpep:
MySQL does not allow primary key fields to be null.  while it silently swallows any request to allow null values, this library ought to point out the failure to users


coleifer:
Cool, thnx!

PR #511, opened 2015-01-22
default values

d1hotpep:
field default values were being ignored for column creations and modifications.  this change ensures that a field marked with a default value creates a default value in the db.  tested this change through the standard Model table creation and schema migrator add_column() function, in mysql and sqlite

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -761,7 +761,9 @@ class Field(Node):
             ddl.append(SQL('NOT NULL'))
         if self.primary_key:
             ddl.append(SQL('PRIMARY KEY'))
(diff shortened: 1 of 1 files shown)


coleifer:
Peewee makes a point of not setting default DB values, so the current behavior is correct.

d1hotpep:
was it too complicated to get right across platforms?

coleifer:
My feeling was that it could lead to confusing behavior on the python side. If you really want it, you can always use SQL to accomplish default DB values, but it wasn't something I wanted to add to peewee.

PR #513, opened 2015-01-26
query results length

d1hotpep:
a couple times i tried to call len() on the results of a query and it exploded...figured this would be good functionality to add.  if you only want a count, i realize that a SELECT COUNT(*) should be used, but there are also times when it would be nice to see how many results i got and use them

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1802,6 +1802,9 @@ class QueryResultWrapper(object):
                 self.next()
             except StopIteration:
                 break
+    
+    def __len__(self):
+        return self.cursor.rowcount
 
 class ExtQueryResultWrapper(QueryResultWrapper):
     def initialize(self, description):
@@ -2556,6 +2559,9 @@ class SelectQuery(Query):
             index += 1
         res.fill_cache(index)
         return res._result_cache[value]
+
+    def __len__(self):
+        return len(self.execute())
 
     if PY3:
         def __hash__(self):


coleifer:
Can you define "exploded"?

coleifer:
There is already the `SelectQuery.count()` method, which handles some edge-cases, so I'm loth to add another method that does the same thing.

d1hotpep:
exploded = throws an exception

>TypeError: object of type 'SelectQuery' has no len()

it just felt like the query results object is an iterable container class that should implement the len() function so you can figure out how many results were returned prior to iterating through them.  otherwise, you have to do something like

```
for res in results:
    do_stuff_with_res(res)
    count += 1
print "you have %d results" % count
```

using Select.count() requires that you either only want a count and not the resulting rows, or else that you have to use two queries.  one snag i found:

>Note that it is not uncommon for databases to report -1 after a select statement for performance reasons. (The exact number may not be known before the first records are returned to the application.) - https://code.google.com/p/pyodbc/wiki/Cursor

this might explain all the Travis failures

coleifer:
There are quite a few test failures resulting from peewee checking the truthiness of the value.

coleifer:
If you want the length of the result set, you can always do ``len(list(query))``.

I think I'm going to pass on merging this and mark wontfix.

yohanboniface:
Why not adding this to SelectQuery:

```
    def __len__(self):
        return self.count()
```

Calling `len` is very pythonic, and is missing imho.

> If you want the length of the result set, you can always do len(list(query)).

Is this issuing a count on db level, or python level?

coleifer:
As I mentioned above, there are places that check the truthiness of a query variable. Adding a `__len__` method would cause the interpreter to use the query's count for the truth check.

This may be worth looking into, though.

coleifer:
I looked again and whatever tests had been failing appear to have been fixed. I've added your changes.

PR #540, opened 2015-02-27
Traverse multiple FKs to the same model with several joins

sangwa:
Consider the following many-to-many relationship:

```python
from peewee import Model, PrimaryKeyField, ForeignKeyField, CharField

class Entity(Model):
    id = PrimaryKeyField()
    name = CharField(unique=True)

class Relation(Model):
    class Meta:
        indexes = ((('parent', 'child'), True),)
    id = PrimaryKeyField()
    parent = ForeignKeyField(Entity, related_name='children')
    child = ForeignKeyField(Entity, related_name='parents')
    data = CharField()
```

If I want to fetch the list of *relations* and access their members of **both** sides ("parent" and "child") without additional queries, for example:
```python
parents = Entity.alias()
children = Entity.alias()
relations = (Relation
                .select(Relation, parents, children)
                .join(parents, on=(Relation.parent == parents.id))
                .switch(Relation)
                .join(children, on=(Relation.child == children.id)))
for rel in relations:
    print rel.id, rel.parent.name, rel.children.name
```
this won't work, because the current implementation of joins will only scan the list of fields up to the **first** foreign key to the joined model, so the second `.join(children, ...)` in the code above will stop scanning on the `parent` attribute and set it second time after the first `.join(parents, ...)` instead of setting the `child` attribute which is actually specified in the expression of the join (and data for it is correctly selected by the generated query).

To overcome this, the patch introduced iterates over all the foreign keys to the same model and selects the appropriate target field based on the join expression, if provided with, so the example code above works properly.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -598,12 +598,12 @@ JoinMetadata = namedtuple('JoinMetadata', (
     'source', 'target_attr', 'dest', 'to_field', 'related_name'))
 
 class Join(namedtuple('_Join', ('dest', 'join_type', 'on'))):
-    def get_foreign_key(self, source, dest):
-        fk_field = source._meta.rel_for_model(dest)
-        if fk_field is not None:
-            return fk_field, False
-        reverse_rel = source._meta.reverse_rel_for_model(dest)
-        if reverse_rel is not None:
+    def get_foreign_keys(self, source, dest):
+        fk_fields = source._meta.all_rels_for_model(dest)
+        if fk_fields:
+            return fk_fields, False
+        reverse_rel = source._meta.all_reverse_rels_for_model(dest)
+        if reverse_rel:
             return reverse_rel, True
         return None, None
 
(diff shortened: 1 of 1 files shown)


coleifer:
The way to do this with peewee currently is to use aliases in the joins, which tell peewee what attribute to patch the related data onto.

If I modify your example,

```python
relations = (Relation
                .select(Relation, parents, children)
                .join(parents, on=(Relation.parent == parents.id).alias('parent'))
                .switch(Relation)
                .join(children, on=(Relation.child == children.id).alias('child')))
```

I can get the related values and ensure that they are being retrieve in a single query:

```python
from playhouse.test_utils import assert_query_count
with assert_query_count(1):
    for rel in relations:
        print rel.id, rel.parent.name, rel.child.name
```

sangwa:
Thanks! Tried aliases before but somehow it didnt work.

Issue #240, opened 2013-10-20
Incompleted transactions with PostgreSQL


rudyryk:
How to deal with incompleted transactions in PostgreSQL? It seems we can't make queries after single transaction fail. Instead we get: **psycopg2.InternalError: current transaction is aborted, commands ignored until end of transaction block**

Here's minimal code sample to reproduce the issue:

``` Python
import peewee
db = peewee.PostgresqlDatabase(
            database='test',
            host='localhost',
            port=None,
            user='test',
            password=None)

class User(peewee.Model):
   username = peewee.CharField(max_length=50, unique=True)

   class Meta:
       database = db

User.create_table()
User.create(username='test')
```

So, I get single user with username "test". If I try to create another one, I'll get an error:

``` Python
>>> User.create(username='test')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 2193, in create
    inst.save(force_insert=True)
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 2278, in save
    new_pk = insert.execute()
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 1623, in execute
    return self.database.last_insert_id(self._execute(), self.model_class)
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 1314, in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 1721, in execute_sql
    res = cursor.execute(sql, params or ())
psycopg2.IntegrityError: duplicate key value violates unique constraint "user_username"
DETAIL:  Key (username)=(test) already exists.
```

And that's predictable and fine! (By the way, how should we catch this kind of exception?)

But when I try to make another request, it fails:

``` Python
>>> User.get(username='test')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 2203, in get
    return sq.get()
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 1521, in get
    return clone.execute().next()
  File "/Users/test/.virtualenvs/test/lib/python3.3/site-packages/peewee.py", line 1556, in execute
[...]

PR #541, opened 2015-03-04
In PostgresqlExtDatabase, respect the 'autorollback' option

davidmcclure:
First of all, thanks so much for Peewee! I ran across the problem described in #240, which was fixed by the addition of the `autorollback` database option.

I was setting `autorollback=True`, but still hitting the problem, and realized that the implementation of `execute_sql ` in `PostgresqlExtDatabase` (which I'm using) doesn't include the check to see if the transaction should be automatically rolled back if the query throws an exception. I just copied in the two lines from the parent `Database` class that do this.


coleifer:
Thanks for this!

PR #544, opened 2015-03-08
Max length reflection

garar:
Hello!

I've added support for max_length to reflection. Previously reflection didn't read max_length for varchar or char.

Left me know if you need anything else for this.

Thanks!


coleifer:
I think I'm going to pass on this, but thank you for the PR.

garar:
Hi!

Can I get some arguments against it? Currently when you run generate_models from Introspector it will return invalid models. max_length with be incorrect for CHAR and VARCHAR fields.

Thanks!

PR #548, opened 2015-03-13
Error with long names in tables 

sidan93:
DATABASE: postgressql
only works postgressql!

I have a table called "TableTableTable" with primary key "@TableTableTable"
CURRVAL complains of a very long name "TableTableTable_@TableTableTable_seq" and cuts it


coleifer:
I think the better fix is to just manually specify the sequence name to be whatever postgresql truncated it to. e.g.:

```python
tabletabletable = PrimaryKeyField(sequence='tabletabletable_id_s')
```

PR #549, opened 2015-03-13
Custom bindings for selection expressions

jhorman:
Currently there is no way to have the results of a selection sub expression land on a joined instance, instead of on the top level instance. Many times a model is expected to always have a field defined, even if it was inflated via a join.

Example:

```python
BlogEntry.select(
    BlogEntry, 
    User, 
    fn.Exists(Role.select(Role.id).where(
        Role.user == User.id,
        Role.role == 'admin'
    )).alias('is_admin').bind_to(User)
).join(User)
```

Here, we expect the `User` objects to always have `is_admin` defined, even if inflated via a join.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -331,6 +331,7 @@ class Node(object):
     def __init__(self):
         self._negated = False
         self._alias = None
+        self._target = None
         self._ordering = None  # ASC or DESC.
 
     @classmethod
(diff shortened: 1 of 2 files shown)


jhorman:
Related to topic here https://groups.google.com/forum/#!topic/peewee-orm/Ny-2UE6yPvs

coleifer:
Neat! Do you mind renaming `target` to `bind_to` and adding documentation on the `Node` class' api docs?

jhorman:
Sounds good. Made those changes @coleifer

PR #559, opened 2015-03-22
Add pwiz option to preserve original DB column order.

elgow:
Add pwiz option to preserve original DB column order in generated model definitions.  Very useful when a DB created by Peewee will be loaded from a dump file created with native DB tools (e.g. Postgresql INSERT type dump).

### DIFF
--- a/playhouse/reflection.py
+++ b/playhouse/reflection.py
@@ -1,4 +1,4 @@
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
 import re
 
 from peewee import *
@@ -135,7 +135,7 @@ class Metadata(object):
         return self.database.execute_sql(sql, params)
 
     def get_columns(self, table, schema=None):
-        metadata = dict(
+        metadata = OrderedDict(
             (metadata.name, metadata)
             for metadata in self.database.get_columns(table, schema))
 
@@ -149,7 +149,7 @@ class Metadata(object):
             if column_types[pk] is IntegerField:
                 column_types[pk] = PrimaryKeyField
 
-        columns = {}
+        columns = OrderedDict()
         for name, column_data in metadata.items():
             columns[name] = Column(
                 name,
--- a/playhouse/reflection.py
+++ b/playhouse/reflection.py
@@ -1,4 +1,9 @@
-from collections import namedtuple, OrderedDict
+from collections import namedtuple
+try:
+    from collections import OrderedDict
+except ImportError:
+    OrderedDict = dict
+
 import re
 
 from peewee import *
--- a/playhouse/reflection.py
+++ b/playhouse/reflection.py
@@ -1,9 +1,9 @@
-from collections import namedtuple
 try:
     from collections import OrderedDict
 except ImportError:
     OrderedDict = dict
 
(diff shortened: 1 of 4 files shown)


elgow:
Hello Charles,

This is a minor modification to enable pwiz to create models that preserve the column ordering from the original source DB. I needed that in order to use dump files created with native Postgresql tools from the original DB. I made the new behavior an option so that default pwiz behavior would be unchanged.  I hope that you will find it to be a useful addition. 

   Ed

elgow:
Hello Charles,

Please drop this PR for now. I've fixed the test problem, but I'm still trying to sort out what to do about OrderedDict for Python 2.6

coleifer:
Nice work. Could you add a unit test?

playhouse/reflection.py
```diff
@@ -1,4 +1,9 @@
 from collections import namedtuple
+try:
+    from collections import OrderedDict
+except ImportError:
+    OrderedDict = dict
+
```
coleifer: Small thing, but could you move the namedtuple import below the try/except? That way they're alphabetized.

pwiz.py
```diff
@@ -138,6 +141,8 @@ def get_option_parser():
     ao('-i', '--info', dest='info', action='store_true',
        help=('Add database information and other metadata to top of the '
              'generated file.'))
+    ao('-o', '--order-preserved', action='store_true',
```
coleifer: Maybe use `dest='preserve_order'` for consistency with the parameter name?

elgow: Hello Charles,

I�ll be happy to oblige on all three of your requested improvements. I�ll send you a new PR when it�s done. 

Thanks again for the INSERT � RETURNING version 2.5.0. It helps me a lot.

  Ed

elgow:
Hi Charles,

All fixes in. Please let me know if you think the way I've divided up the EXPECTED text is "too clever by half" and if you'd prefer just plain text for the new test. 

  Ed

coleifer:
> Please let me know if you think the way I've divided up the EXPECTED text is "too clever by half" and if you'd prefer just plain text for the new test.

errr yeah

elgow:
Hello Charles,

I'm at a loss to explain the unit test failures for the preserve_order test.  It appears that in the note table the columns are being defined in the reverse of the order in which they are listed in the class definition. This does not happen on either of my machines. The column creation order is controlled by this code 

        # Used internally for recovering the order in which Fields were defined
        # on the Model class.
        Field._field_counter += 1
        self._order = Field._field_counter
        self._sort_key = (self.primary_key and 1 or 2), self._order

so that the PK columns should sort first and then others will follow, all in the order in which they were declared. The test output seems to indicate that this is not happening. Could you try running the tests outside of the CI system to see what result you get? I have not been able to reproduce the test failure. 

Thanks,
   Ed

elgow:
I switched to full explicit text for each expected value per your preference. Also prevented test failure of unsupported feature under Python 2.6

coleifer:
Thank you so much! I've mergd a very slightly modified version of your changes. I also caught an unreported bug while messing with the tests, so even better!

PR #561, opened 2015-03-25
insert_many() + insert_returning support

ianawilson:
I saw that there was support for RETURNING but only for single-row INSERTs, and I needed support for multi-row INSERTs with RETURNING as well.

Based on your tests, it looks like you may not have wanted to include this functionality. If that's the case, I'd be happy to make this a non-default option for `insert_many()` so that the default behavior is unchanged, but getting the primary keys of all new rows is available as an option for those who want it.

Otherwise, if this does seem like a good default behavior, then I've updated the tests and this should be good to merge.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1676,7 +1676,7 @@ class QueryCompiler(object):
                     SQL('VALUES'),
                     CommaClause(*value_clauses)])
 
-        if meta.database.insert_returning and not query._is_multi_row_insert:
+        if meta.database.insert_returning:
             clauses.extend([
                 SQL('RETURNING'),
                 self._get_field_clause(
@@ -2760,6 +2760,8 @@ class InsertQuery(Query):
             else:
                 return self.database.last_insert_id(cursor, self.model_class)
         else:
+            if self.database.insert_returning:
+                return [row[0] for row in cursor.fetchall()]
             return True
 
 class DeleteQuery(Query):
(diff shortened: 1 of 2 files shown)


coleifer:
I didn't think it made a lot of sense to return the IDs for bulk inserts. If inserting a lot of rows, it could mean a lot of data being transferred back over the wire for all those inserts. Furthermore, insert_many() doesn't return instances, so what would be the point of having the IDs?

Also, the behavior prior to 2.5 was to only return the latest insert ID of a batch.

ianawilson:
In my particular use case, I'm using peewee in a service that ingests large amounts of data and writes it to the db (postgres). We have another service that is responsible for updating the search index (elasticsearch), and it drastically simplifies things if I can pass the list of new IDs to the search index service after the bulk insert.

I'm not convinced this behavior makes sense as a default, but I do think it's nice to have as an option, especially because the behavior in 2.5 is nearly there. Like I mentioned, I'd be happy to rewrite it to be a non-default option for `insert_many()` / `InsertQuery` -- shouldn't be hard.

coleifer:
> I do think it's nice to have as an option

Sounds good to me.

coleifer:
You can now call:

```python
data = [{'username': username} for username in list_of_usernames]
user_ids = User.insert_many(data).return_id_list().execute()
print user_ids
```

coleifer:
http://docs.peewee-orm.com/en/latest/peewee/api.html#InsertQuery.return_id_list

PR #564, opened 2015-03-27
Update pk seq

elgow:
New method to force the DB to ensure that the next auto-generated key will be greater than any existing key, including explicitly inserted keys. On sqlite and MySQL this happens automatically, so the method is a no-op. For Postgresql it provides a canned version of a nice query to update the sequence, saving peewee users from having to write it themselves. 

This might possibly be simplified and better integrated by having a method to turn auto_increment on/off rather than just setting the attribute on the model. Then this method could be run whenever auto_increment is explicitly activated.

### DIFF
--- a/docs/peewee/models.rst
+++ b/docs/peewee/models.rst
@@ -478,6 +478,7 @@ Sometimes you do not want the database to automatically generate a value for the
             u.save(force_insert=True) # <-- force peewee to insert row
 
     User._meta.auto_increment = True
+    User.update_auto_pk()  # <--- force DB to ensure that next generated key will be > any existing key
 
 If you *always* want to have control over the primary key, simply do not use the :py:class:`PrimaryKeyField` field type, but use a normal :py:class:`IntegerField` (or other column type):
 
--- a/peewee.py
+++ b/peewee.py
@@ -3080,6 +3080,10 @@ class Database(object):
     def truncate_date(self, date_part, date_field):
         return fn.DATE_TRUNC(SQL(date_part), date_field)
 
+    def update_auto_pk(self):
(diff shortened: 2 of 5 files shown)


coleifer:
Thank you for the effort you put into this, but I don't have any plans to merge this functionality, as it seems to me something that would be a better fit for your own library code.

elgow:
I can certainly just use it myself, but I thought it might be helpful for the people who follow the peewee docs in the section "Manually specifying primary keys".  After they've re-activated auto_increment in a Postgresql database, as instructed by the docs, then they are going to encounter a duplicate key constraint violation unless they do something like what I've done.

I'm aware that this operation is perilous in a highly concurrent environment, but for populating initial values for drop-downs, etc. into tables, it could be useful, especially for DB novices starting out with peewee.

coleifer:
> The use case for this is actually very common

I'd beg to differ, as in the 4+ years peewee's been around, you're the first to ask for a sequence updating helper.

My stance is that relying on particular values for auto-incremented, sequence-backed columns is an anti-pattern. I understand that maybe you're restoring data, but in the very rare cases I've done that all I had to do was restore the data then set the currval of the sequence -- it was a one-time operation, just another part of restoring from backup.

The docs you've referenced even emphasize that this is a one-time type operation:

> To handle this on a **one-off** basis, you can simply tell peewee to turn off auto_increment during the import

elgow:
I agree that it's a one-off operation, though possibly one time for each new deployment. But when you are finished with it the DB might be broken (if it's postgresql), or it might not. The docs don't mention that your DB might be broken, and there's no database neutral code that you can write to make sure that it's fixed. That's all I'm trying to create, the simplest possible uniform way to fix any DB after inserting explicit keys into an auto-increment table.

PR #569, opened 2015-04-02
Expose parsed db url

stt:
It'd be useful to get the parsed dict from db_url so it could be used with pooled connections as well.
E.g.:

	db = PooledMySQLDatabase(
	    max_connections=32,
	    stale_timeout=300,  # 5 minutes.
	    **db_url.parse(app.config['DATABASE_URL'])
	    )

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -31,17 +31,7 @@ schemes = {
     'sqliteext': SqliteExtDatabase,
 }
 
-def connect(url):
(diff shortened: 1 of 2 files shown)


coleifer:
Nice! Do you mind adding a test-case for the new function?

stt:
Sure, :+1: ?

coleifer:
Thanks!

PR #578, opened 2015-04-22
let the Meta class can extends from another class

anjianshi:
Example:
```python
db = PostgresqlDatabase("mydb")

class BaseMeta:
    database = db

class BaseModel(Model):
    class Meta(BaseMeta):
        pass

class ModelA(Model): 
    pass

class ModelB(Model):
    class Meta(BaseMeta):
        db_table = "model_b"
```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -3797,9 +3797,9 @@ class BaseModel(type):
         meta_options = {}
         meta = attrs.pop('Meta', None)
         if meta:
-            for k, v in meta.__dict__.items():
+            for k in dir(meta):
                 if not k.startswith('_'):
-                    meta_options[k] = v
+                    meta_options[k] = getattr(meta, k)
 
         model_pk = getattr(meta, 'primary_key', None)
         parent_pk = None


coleifer:
I can see the reasoning behind this change, but I'm loth to introduce any changes that might break existing applications. I'm not sure off-hand, but given the way peewee munges `Meta` objects at present, there may be some edge-cases or situations where this breaks.

What do you think?

anjianshi:
I see, the `dir()` function may be unstable.
Can we use `inspect.getmembers()` instead? 
http://stackoverflow.com/a/8529470/2815178
I think it's more safety.
```python
import inspect
for k, v in inspect.getmembers(meta):
    if not k.startswith('_'):
        meta_options[k] = v
```

anjianshi:
Or I can just use a decorator:

```
def with_database(cls):
    cls.database = db
    return cls


class BaseModel(Model):
    @with_database
    class Meta:
        pass

class ModelA(Model): 
    pass

class ModelB(Model):
    @with_database
    class Meta:
        db_table = "model_b"
```

Then, nothing need changes in Peewee.
It's just not so nature.

coleifer:
You do know you can write this, right?

```python
db = PostgresqlDatabase("mydb")

class BaseModel(Model):
    class Meta:
        database = db

class ModelA(BaseModel): 
    # no need to define Meta / database = db, it is inherited.
    pass

class ModelB(BaseModel):
    class Meta(BaseMeta):
        # database = db --> this is inherited via BaseModel
        db_table = "model_b"
```

anjianshi:
Oh, I don't know this before. Thank you!
Perhaps this pull request is not needed any more?

PR #591, opened 2015-05-04
fix when not using named cursor

eseom:
I'm using peewee on my projects. Thanks. :)

I wrote a simple connection to the postgresql server without named cursor.
```
from peewee import *  # NOQA
from playhouse.postgres_ext import *  # NOQA

database = PostgresqlExtDatabase('test', user='test')


class MyModel(Model):
    name = CharField()
    value = CharField()

    class Meta:
        database = database


MyModel.create_table()
```
An error occurred.
```
  File "test.py", line 15, in <module>
    MyModel.create_table()
  File "/home/vagrant/src/peewee/peewee.py", line 4010, in create_table
    db.create_table(cls)
  File "/home/vagrant/src/peewee/peewee.py", line 3115, in create_table
    return self.execute_sql(*qc.create_table(model_class, safe))
  File "/home/vagrant/src/peewee/playhouse/postgres_ext.py", line 356, in execute_sql
    cursor = self.get_cursor()
  File "/home/vagrant/src/peewee/playhouse/postgres_ext.py", line 343, in get_cursor
    return self.get_conn().cursor(name=name)
TypeError: argument 1 must be string, not None
```

### DIFF
--- a/playhouse/postgres_ext.py
+++ b/playhouse/postgres_ext.py
@@ -340,7 +340,9 @@ class PostgresqlExtDatabase(PostgresqlDatabase):
         super(PostgresqlExtDatabase, self).__init__(*args, **kwargs)
 
     def get_cursor(self, name=None):
-        return self.get_conn().cursor(name=name)
+        if name:
+            return self.get_conn().cursor(name=name)
+        return self.get_conn().cursor()
 
     def execute_sql(self, sql, params=None, require_commit=True,
                     named_cursor=False):


coleifer:
Interesting, what version of psycopg2 are you using?

eseom:
missing my environment
ubuntu 14.04
python 2.7.6
psycopg2 is 2.4.5 from python-psycopg2 package

coleifer:
Ew, just a tip but you should definitely be using pip and not the Ubuntu psycpog2 package.

That said, seems like a fine PR so I'll go ahead and merge.

PR #607, opened 2015-05-20
Introspector doing too much work

pilate:
Tables from arguments aren't being passed to introspector.


coleifer:
Nice catch, thanks!

PR #610, opened 2015-05-23
Added id accessor to foreign relations, similar to Django.

tals:
Models with a ForeignKeyField instance now receive an additional attribute - attr_name_id. This allows you to access the foreign id without bringing in the entire model.

It's currently missing tests, but I figured I should see if you're interested in this first :)


coleifer:
Well, I originally didn't want to add this, but thinking about it, enough people have requested this that maybe it'd be a good idea. Thanks I'll go ahead and add some tests.

PR #622, opened 2015-06-09
Support connect params in db_url

klen:
Hello!

The PR provides a way to set additional connection params when db_url.connect is used. Such as `encoding` for postgresql and etc.


coleifer:
Beautiful!

PR #627, opened 2015-06-15
Migrations: change column

klen:
I've used the functionality in my project with Postgres and it works for me. So I thought it could be useful for other people. The request provides beta support for `change_column` functionality in migrator. I've tested this with Postgres and SQLite.

### DIFF
--- a/playhouse/migrate.py
+++ b/playhouse/migrate.py
@@ -227,6 +227,19 @@ class SchemaMigrator(object):
         return operations
 
     @operation
+    def change_column(self, table, column_name, field):
+        operations = [self.alter_change_column(table, column_name, field)]
+        if not field.null:
+            operations.extend([self.add_not_null(table, column_name)])
+        return operations
+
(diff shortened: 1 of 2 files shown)


coleifer:
I think I will pass on this, but thank you.

klen:
What wrong?

coleifer:
Well, I think `ALTER TABLE / ALTER COLUMN` is just complicated enough that I'm not sure this implementation will suffice. Additionally, the MySQL syntax is quite a bit different IIRC, and is not added as part of this PR.

klen:
That the reason why I didn't include the functionality in docs and chosen such word as `beta`. But the code (uncompleted of course) already helped me with postgres and sqlite.

But it up to you, coz. Thank you for the project!

coleifer:
Thanks for the PR and for taking the time to submit the code. If I should decide to add this functionality in the future I will definitely use this PR as the starting place.

PR #637, opened 2015-06-24
Added small bit in documentation about ForeignKeyField's related_name

josefdlange:
I had a lengthy argument with a co-worker about the implicit `related_name` backreference created by peewee when declaring a `ForeignKeyField`. He was convinced it didn't exist, since it wasn't in the documentation. Only after walking him through the source code and demonstrating it did he believe me. So, for the future, I added a small section in the `ForeignKeyField` docs explaining. 

I do not know how to build/test `peewee`'s documentation but would be glad to if I had some direction. Also accepting changes to what I wrote, if it doesn't match up with the original intent or design.

Also added a `.gitignore` to ignore `.pyc` files and the `build` directory. We can trim that out of here if you feel it doesn't belong, though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+*.pyc
+build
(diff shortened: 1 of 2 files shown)


coleifer:
Sick, I'll take the gitignore too!

PR #645, opened 2015-07-04
Make filter work when there are multiple joins to the same table

rammie:


coleifer:
This is some pretty dense code... Can you summarize the changes? I'm not sure I'm going to merge it in yet, but some test-cases would also be helpful for my understanding of these changes.

peewee.py
```diff
@@ -2437,15 +2446,31 @@ def filter(self, *args, **kwargs):
 
         dq_node = dq_node.rhs
 
+        seen = set()
+        remove_dupes = lambda lst: seen.clear() or [
+            i for i in lst if i not in seen and not seen.add(i)]
```
coleifer: `set.clear()` and `set.add()` seem to always return `None`. What's going on here?

coleifer:
Actually...man, this code is very difficult to parse in my head. Might be the 3 hours of sleep I'm running on, but since `filter()` is deprecated anyways I'm not too excited about adding this much complexity and then having to maintain it.

rammie:
Let me take a stab at cleaning it up, and I'll add some tests as well. As I mentioned before, I like the way that the flask-peewee REST API processes the query string to generate a query, which relies on ```filter()```. Maybe this could be moved to the playhouse instead?

peewee.py
rammie: It clears the set before returning the result of the list comprehension. The returned list does not contain any duplicate items. I could move this to a separate method and out of a list comprehension.

http://stackoverflow.com/questions/7961363/python-removing-duplicates-in-lists

rammie:
On second thought, I decided to move this to the REST base classes itself API. Seems more logical if you're deprecating ```filter()``` anyway.

Issue #647, opened 2015-07-07
Introducing returning for postgres


IxDay:
Hello, I am currently using peewee for managing a postgresql and I have started implementing a returning function on the update and select queries.

I am interested in contributing the results.

Here is an example:

```python
user = User.update(active=False).where(User.registration_expired == True).returning()
```

The returning method will be a SelectQuery and support the selection arguments

```python
sq = (User.update(active=False)
          .where(User.registration_expired == True)
          .returning(User.username))
for user in sq:
    print user.username
```

And also support the `tuples` and `dicts` chaining:

```python
sq = (User.update(active=False)
          .where(User.registration_expired == True)
          .returning(User.username)
          .dicts())
for user in sq:
    print user.get('username')
```

coleifer:
Interesting, how is this implemented at the SQL level?

IxDay:
http://www.postgresql.org/docs/9.1/static/sql-update.html

`If the UPDATE command contains a RETURNING clause, the result will be similar to that of a SELECT statement containing the columns and values defined in the RETURNING list, computed over the row(s) updated by the command` (the last statement is false, every row can be accessed I tested it) .

The sql statement will look like this

```sql
UPDATE users SET active = FALSE WHERE registration_expired = TRUE RETURNING username; 
```

coleifer:
Gotcha, cool! Can you share your code?

IxDay:
I will make a first pull request in order to see if the implementation looks good for you, with a simple test case.

coleifer:
I've gone ahead and merged support for `RETURNING` clauses on `UPDATE`, `INSERT` and `DELETE` queries. These only work with Postgresql, and are protected using a database config attribute, `returning_clause`. I will be adding documentation as well today, then I think it's good to go!

PR #649, opened 2015-07-07
Work in progress

IxDay:
Here is a first implementation just showing part of the code.

I have to fix the ability to __iter__ through results
I also have an issue with the model class, If you have a better idea than creating a dedicated model class...

It breaks the tests but this is a work in progress ;)

linked to #647


IxDay:
Wrong repository

PR #650, opened 2015-07-07
Work in progress

IxDay:
Here is a first implementation just showing part of the code.

I have to fix the ability to iter through results
I also have an issue with the model class, If you have a better idea than creating a dedicated model class...

It breaks the tests but this is a work in progress ;)

linked to #647

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4071,6 +4071,13 @@ class BaseModel(type):
         cls.DoesNotExist = exception_class
         cls._meta.prepared()
 
+        import playhouse.postgres_ext as postgres_ext
+        db = cls._meta.database
+        if isinstance(db, postgres_ext.PostgresqlExtDatabase):
+            cls._update_cls = postgres_ext.PostgresqlExtUpdateQuery
+        else:
+            cls._update_cls = UpdateQuery
+
         return cls
 
     def __iter__(self):
@@ -4099,7 +4106,7 @@ class Model(with_metaclass(BaseModel)):
     @classmethod
     def update(cls, **update):
         fdict = dict((cls._meta.fields[f], v) for f, v in update.items())
-        return UpdateQuery(cls, fdict)
+        return cls._update_cls(cls, fdict)
 
     @classmethod
     def insert(cls, **insert):
--- a/playhouse/postgres_ext.py
+++ b/playhouse/postgres_ext.py
@@ -5,6 +5,7 @@ Collection of postgres-specific extensions, currently including:
 """
 import uuid
 
+import peewee
 from peewee import *
 from peewee import Expression
 from peewee import logger
@@ -275,6 +276,7 @@ OP.update(
 
 
 class PostgresqlExtCompiler(QueryCompiler):
+
     def _create_index(self, model_class, fields, unique=False):
         clause = super(PostgresqlExtCompiler, self)._create_index(
             model_class, fields, unique)
@@ -330,6 +332,19 @@ class PostgresqlExtCompiler(QueryCompiler):
             json_path=self._parse_json_path)
         return parse_map
 
+    def generate_update(self, query):
+        sql, params = super(PostgresqlExtCompiler, self).generate_update(query)
+        if query._returning:
+            clauses = [peewee.SQL(' RETURNING')]
+            returning_clause = peewee.Clause(*query._returning)
+            returning_clause.glue = ', '
+            clauses.append(returning_clause)
+            returning_sql, _ = self.build_query(clauses)
+            sql += returning_sql
+
+        return sql, params
+
+
 
 class PostgresqlExtDatabase(PostgresqlDatabase):
     compiler_class = PostgresqlExtCompiler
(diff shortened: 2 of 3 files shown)


IxDay:
Just before continuing do you think that I should continue implementing a custom Model class or you have a better idea on how to implement this?

I will also look at the implementation of the InsertQuery and avoid code duplication.

coleifer:
I'll take a look soon. One thing, you can probably just modify the existing `UpdateQuery` rather than putting this in `postgres_ext`. That's how I handled `INSERT...RETURNING` in the existing code.

peewee.py
```diff
@@ -4071,6 +4071,13 @@ def __new__(cls, name, bases, attrs):
         cls.DoesNotExist = exception_class
         cls._meta.prepared()
 
+        import playhouse.postgres_ext as postgres_ext
+        db = cls._meta.database
+        if isinstance(db, postgres_ext.PostgresqlExtDatabase):
+            cls._update_cls = postgres_ext.PostgresqlExtUpdateQuery
+        else:
+            cls._update_cls = UpdateQuery
+
```
coleifer: This feels really hacky to me.

coleifer:
I might try my hand at an implementation -- by doing so I'd like to explore some other ways of implementing this.

coleifer:
This is what I've come up with: https://github.com/coleifer/peewee/compare/HEAD...feature/update-returning

playhouse/postgres_ext.py
```diff
@@ -330,6 +343,18 @@ def get_parse_map(self):
...
 
+    def generate_update(self, query):
+        sql, params = super(PostgresqlExtCompiler, self).generate_update(query)
+        if getattr(query, '_returning', None):
+            clauses = [SQL(' RETURNING')]
+            returning_clause = Clause(*query._returning)
+            returning_clause.glue = ', '
+            clauses.append(returning_clause)
+            returning_sql, _ = self.build_query(clauses)
+            sql += returning_sql
```
coleifer: Note that you are not appending any parameterized values from the `RETURNING` clause to the list of values from the main query.

playhouse/postgres_ext.py
```diff
@@ -376,6 +401,80 @@ def _connect(self, database, **kwargs):
...
+            """Add a tuples wrapper"""
+            self._tuples = tuples
+
+        def dicts(self, dicts=True):
+            """Add a dicts wrapper"""
+            self._dicts = dicts
+
+        setattr(PostgresqlExtUpdateQuery, 'naive', returns_clone(naive))
+        setattr(PostgresqlExtUpdateQuery, 'tuples', returns_clone(tuples))
+        setattr(PostgresqlExtUpdateQuery, 'dicts', returns_clone(dicts))
```
coleifer: Why aren't these just regular methods on the class?

playhouse/postgres_ext.py
```diff
@@ -376,6 +401,80 @@ def _connect(self, database, **kwargs):
...
+
+    def _clone_attributes(self, query):
+        """Clone the specific attributes for chaining"""
+        query = super(PostgresqlExtUpdateQuery, self)._clone_attributes(query)
+        query._returning = self._returning
```
coleifer: I also cloned `._tuples` and `._dicts` in my code.

playhouse/postgres_ext.py
```diff
@@ -376,6 +401,80 @@ def _connect(self, database, **kwargs):
...
+
+        if self._dirty or not self._qr:
+            if self._tuples:
+                ResultWrapper = TuplesQueryResultWrapper
+            elif self._dicts:
+                ResultWrapper = DictQueryResultWrapper
+            elif self._naive:
+                ResultWrapper = NaiveQueryResultWrapper
+            else:
+                ResultWrapper = ModelQueryResultWrapper
```
coleifer: Since joins are not allowed on UPDATE queries with Peewee, I'm not sure of the case where you'd want the `ModelQueryResultWrapper` over the `NaiveQueryResultWrapper`.

playhouse/postgres_ext.py
```diff
@@ -376,6 +401,80 @@ def _connect(self, database, **kwargs):
...
+
+            meta = self.get_query_meta()
+            self._qr =  ResultWrapper(self.model_class, self._execute(), meta)
+            self._dirty = False
```
coleifer: What would be some example use-cases where having the `_dirty` attribute would be necessary?

coleifer:
I noticed that you put all this code in the `postgresql_ext` module. My preference, since I imagine *some* DBs besides postgres have this feature, is to use database feature flags to allow/disallow access to the `returning` APIs.

My thoughts on the implementation can be found here: https://github.com/coleifer/peewee/compare/HEAD...feature/update-returning

coleifer:
You know, looking around it seems like `DELETE` and `INSERT` both support `returning` as well. I think I'll update my patch to allow `returning` on delete and insert.

coleifer:
I've added support for `RETURNING` with `UPDATE`, `INSERT` and `DELETE` in my branch `feature/update-returning`:

https://github.com/coleifer/peewee/compare/HEAD...feature/update-returning

I think I'm going to go ahead and merge it today.

IxDay:
Just saw your patch (was on holidays), seems really cool, thanks for the quick implementation :)

coleifer:
Cool, thanks for sharing your code, really helpful!

PR #670, opened 2015-07-25
Minor docs fix regarding Django middleware

messense:
Django middleware ``process_response`` method must return an ``HttpResponse`` or ``StreamingHttpResponse`` object as its documentation denotes.

https://docs.djangoproject.com/en/1.8/topics/http/middleware/#process-response


coleifer:
Thanks!

PR #681, opened 2015-08-03
Add multi-column unique constraints support for Model.create_or_get

messense:


peewee.py
```diff
@@ -4276,11 +4276,27 @@ def create_or_get(cls, **kwargs):
...
         except IntegrityError:
-            query = []  # TODO: multi-column unique constraints.
+            query = []
             for field_name, value in kwargs.items():
                 field = cls._meta.fields[field_name]
                 if field.unique or field.primary_key:
                     query.append(field == value)
+            # Use multi-column unique indexes when
+            # we got no unique/primary_key fields
+            if not query and cls._meta.indexes:
```
foxx: I'm not sure this is necessary, if we have compound unique indexes then they should probably added to the query regardless of whether any other indexes exist. Thoughts @coleifer ?

foxx:
Could probably tidy up the code a little, and we'll need sign off from @coleifer on the `query` conditional, but assuming those two things are signed off, then LGTM.

foxx:
@messense Has this already been merged into master?

messense:
I don't know. I haven't used peewee for a long time and I think I won't be working on this so I closed it.

foxx:
@coleifer Any thoughts on whether you'd like this in master?

coleifer:
No, get or create is evil. One of my biggest regrets

Issue #684, opened 2015-08-08
Does peewee support MySQL REPLACE INTO syntax?


messense:
I only see upsert support for SQLite in documentation.

And another question, does peewee support MySQL ``INSERT IGNORE`` or ``INSERT...ON DUPLICATE KEY UPDATE`` syntax?

coleifer:
Not at present, but I would consider a pull request and would work with you on getting the code into good shape. For now I'd suggest using hand-written SQL queries if you want upserts with MySQL.

PR #687, opened 2015-08-14
Make it possible to use ``upsert`` and ``on_conflict`` on MySQL

messense:
Ref #684

I am not sure how to write a corresponding test case for now.

And I don't know if it is ok to use ``isintance(meta.database, MySQLDatabase`` here. What do you think?


coleifer:
Cool, this is a nice start.

coleifer:
This behavior was added in #753.

PR #699, opened 2015-09-02
djpeewee: Use `model_field.column` instead of `get_attname()`

### LINKED ISSUE #698: djpeewee: ForeignKeys with non-standard column names doesn't work
Having a ref that doesn't match the "<thing>_id" works badly when translating Django models to Peewee.

simonpercivall:
This bacame an issue when the column was named differently than the
standard "<thing>_id".

Fixes #698


coleifer:
Nice, thanks!

PR #700, opened 2015-09-03
Add sensible default from Field if Database() wasn't initialized with fields argument

bdpdx:
Modified QueryCompiler.get_column_type to return an uppercase version of the f argument if the field type doesn't exist in _field_map.

This enables a sensible default if a Field subclass defines a type supported by the underlying database that isn't passed into the Database() constructor via its fields initializer.


coleifer:
Cool, let's try it out!

PR #713, opened 2015-09-15
- Fixed column_name_re not working with ` characters.

### LINKED ISSUE #712: SQLite migration cannot handle all CREATE TABLE queries
I was implementing the use of the `drop_column` method in the `SqliteMigrator` when I found a bug.

I tested the result by changing my database model using sqlitebrowser. The `CREATE TABLE` query sqlitebrowser  generates, contains newlines and \` characters around the table name. This differs from the peewee `CREATE TABLE` queries and the migration code does not support this.

sroebert:
This fixes #712, where migration of a sqlite database is not working when the `CREATE TABLE` query contains newlines or \`.


coleifer:
Thank you!

PR #721, opened 2015-09-24
Make array field adapter independent from the connection.

stas:
Since the field does not depend on the database OIDs, calling `prepare()` is not necessary. This is also mentioned in the docs[1]. The reason this is not dependent on the connection since we don't try to adapt any special OIDs that can vary from one Postgres version to another.

We discovered this when a wrapped connection was used to adapt the new type.
Relevant: opbeat/opbeat_python#68

Thanks in advance for reviewing this.

[1]: http://initd.org/psycopg/docs/extensions.html#psycopg2.extensions.ISQLQuote.prepare

### DIFF
--- a/playhouse/postgres_ext.py
+++ b/playhouse/postgres_ext.py
@@ -107,9 +107,7 @@ class _Array(Node):
         super(_Array, self).__init__()
 
 def adapt_array(arr):
-    conn = arr.field.model_class._meta.database.get_conn()
     items = adapt(arr.items)
-    items.prepare(conn)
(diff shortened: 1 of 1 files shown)


stas:
Hmm, it looks like there's something strange with the way py3 handles the unicode string adapter.

This could be related to psycopg/psycopg2#331
I've tried a couple of things, including calling `adapted.prepare(conn)` and explicitly setting the connection encoding to `UTF8`, still on py3 I get the adapter result as bytea.

We can close this until the py3 support is not ready... :|

PR #735, opened 2015-10-21
adding a .collate_nocase() query feature

aerickson:
Please let me know if you'd like me to make test and/or documentation changes.

Usage example:

```
order_by(Object.name).collate_nocase()
```


aerickson:
I guess this is unnecessary and only works on Sqlite.

fn.Lower() seems to work fine.

```
order_by(fn.Lower(Object.name))
```

coleifer:
Yeah you can also do:

.order_by(Clause(Model.field, SQL(' collate nocase')))

PR #743, opened 2015-11-01
Foreign key to the composite primary key

Maillol:
Create unit test in order to reproduce bug when we create foreign key to the
composite primary key.

### DIFF
--- a/playhouse/tests/models.py
+++ b/playhouse/tests/models.py
@@ -243,6 +243,10 @@ class CompositeKeyModel(TestModel):
         primary_key = CompositeKey('f1', 'f2')
 
 
+class ForeignKeyToCompositeKeyModel(TestModel):
+    ref_to_composite_key_model = ForeignKeyField(CompositeKeyModel)
+    tag = CharField()
+
 class UserThing(TestModel):
     thing = CharField()
     user = ForeignKeyField(User, related_name='things')
@@ -405,4 +409,5 @@ MODELS = [
     SpecialComment,
     EmptyModel,
     NoPKModel,
+    ForeignKeyToCompositeKeyModel
 ]
--- a/playhouse/tests/test_keys.py
+++ b/playhouse/tests/test_keys.py
@@ -289,6 +289,50 @@ class TestCompositePrimaryKey(ModelTestCase):
             ['t1', 't2', 't3'])
 
 
+class TestForeignKeyToCompositePrimaryKey(ModelTestCase):
+    requires = [ForeignKeyToCompositeKeyModel, CompositeKeyModel]
+
+    def setUp(self):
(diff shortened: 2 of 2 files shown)


coleifer:
I've waffled about this some. I dislike implicitly creating the columns. Additionally, when querying, things seem like they might get a little weird, e.g.:

```python
class Person(Model):
    first_name = CharField()
    last_name = CharField()
    class Meta:
        primary_key = CompositeKey('first_name', 'last_name')

# API from your PR:
class Pet(Model):
    owner = ForeignKeyField(Person)

# API I had considered:
class Pet(Model):
    owner_first_name = CharField()
    owner_last_name = CharField()
    owner = CompositeForeignKey(Person)
```

Each has its downsides. In the former, the related person's first and last name, which are present on the Pet, are abstracted away and not easily accessible. In the latter, I'm left wondering how useful that `owner` attribute is really going to be.

Maillol:
I agree with you, creating columns must be explicit.

currently the problem appear when we want create a weak entity  which contains an other weak entity. 
i.e:
Book -> Chapter -> Paragraph 
or
Train -> Wagon -> Seat

The PK Wagon is (PK train + num Wagon) and PK Seat is (PK train + num Wagon + num Seat).
The Wagon num field isn't unique, I can't create a foreign key on it, I must use composite FK. 

What do you think about API like this ?

```python
class Pet(Model):
    owner_first_name = ForeignKey(Person, first_name)
    owner_last_name = ForeignKey(Person, last_name)

    class Meta:
        foreign_keys = [
            CompositeKey(owner_first_name, owner_last_name)
        ]
```

coleifer:
> What do you think about API like this ?

That would be confusing I think. By making `owner_first_name` and `owner_last_name` ForeignKeyFields, it is unclear what value you are supposed to assign to them. Typically with a foreign key field you would assign the related model instance (although it is possible to assign the underyling scalar key value). This "overloads" the concept of foreign keys in a way I find ambiguous.

If there were going to be an API, it would be this most likely:

```python
class Pet(Model):
    owner_first_name = CharField()
    owner_last_name = CharField()
    owner = CompositeForeignKey(Person)
```

But as I said I have doubts about the utility of this.

I think that for now I am going to close this as *wontfix*, since there doesn't seem to me to be a clear path forward. You can specify table-level constraints as model meta options, so if you want schema-level referential integrity, you can do something like:

```python
class Pet(Model):
    owner_first_name = CharField()
    owner_last_name = CharField()

    class Meta:
        constraints = [SQL('FOREIGN KEY(owner_first, owner_last) REFERENCES person(first, last)')]
```

coleifer:
Documentation: http://docs.peewee-orm.com/en/latest/peewee/models.html#indexes-and-constraints

PR #750, opened 2015-11-06
Add connection management documentation for Falcon

arpankapoor:
Middleware components in Falcon framework (http://falconframework.org/)
can be used to add database connection management.

### DIFF
--- a/docs/peewee/database.rst
+++ b/docs/peewee/database.rst
@@ -713,6 +713,28 @@ The connection handling code can be placed in a `middleware <https://pythonhoste
 
 Thanks to GitHub user *@tuukkamustonen* for submitting this code.
 
+Falcon
+^^^^^^
+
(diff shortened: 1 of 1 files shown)


coleifer:
Does the response middleware method not need to return anything? (Or the request middleware for that matter)

arpankapoor:
Nothing needs to be returned.

The middleware methods may modify the request and response objects passed in as parameters. These objects are routed to the appropriate responder method by Falcon.

See the docstring of the middleware methods [here](https://falcon.readthedocs.org/en/latest/api/middleware.html#middleware-components).

coleifer:
Thanks!

arivero:
Not sure if it should be documented, but calling db.close in the process_response kills the cursor, and then a response via stream will fail

PR #757, opened 2015-11-17
Pwiz preservation of composite indices and uniques

SarahFowler73:
When pwiz creates your model, it will preserve the composite indices and composite unique constraints that you have on your tables in your model.

### DIFF
--- a/pwiz.py
+++ b/pwiz.py
@@ -74,6 +74,9 @@ def print_models(introspector, tables=None, preserve_order=False):
         if not preserve_order:
             columns = sorted(columns)
         primary_keys = database.primary_keys[table]
+        composite_indices = [index_data for index_data in database.indexes[table] if len(index_data.columns) > 1]
+        composite_columns = [col for idx in composite_indices for col in idx.columns]
+
         for name, column in columns:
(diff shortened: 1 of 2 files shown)


pwiz.py
```diff
@@ -74,6 +74,9 @@ def _print_table(table, seen, accum=None):
         if not preserve_order:
             columns = sorted(columns)
         primary_keys = database.primary_keys[table]
+        composite_indices = [index_data for index_data in database.indexes[table] if len(index_data.columns) > 1]
+        composite_unq_cols = [col for idx in composite_indices for col in idx.columns if idx.unique]
+        composite_nonunq_cols = [col for idx in composite_indices for col in idx.columns if not idx.unique]
```
coleifer: These lines look to be > 79 characters. Additionally nested list comprehensions are typically something I prefer to avoid in my code.

coleifer: I think this logic should probably live in the introspection module.

coleifer: And actually, it looks like it's there: https://github.com/coleifer/peewee/blob/master/playhouse/reflection.py#L527-L536

PR #768, opened 2015-11-23
Fix _SortedFieldList, Model._meta.remove_field

klen:
Hello, the last release has broken migrations, because ``Model._meta.remove_fields`` actually doesn't work. 

The PR fixes it. I'm waiting for your response and new release.


coleifer:
Thank you!

PR #795, opened 2015-12-17
Allow to pass ids to m2m.add and m2m.remove methods

yohanboniface:


playhouse/fields.py
```diff
@@ -186,7 +186,8 @@ def add(self, value, clear_existing=False):
                 value = [value]
             inserts = [{
                 fd.src_fk.name: self._instance.get_id(),
-                fd.dest_fk.name: rel_instance.get_id()}
+                fd.dest_fk.name: rel_instance.get_id()
+                if hasattr(rel_instance, 'get_id') else rel_instance}
```
yohanboniface: Not a big fan of the indentation here, but at this point of imbrication I'm not sure we can end up with something elegant. :/
One thought: if we had a 'pk' like property, it would help in this kind of situation, doing something like:
```
 fd.dest_fk.name: getattr(rel_instance, 'pk', rel_instance)
```

coleifer:
I opted for a slightly different route for getting the ID list. Thanks for the diff.

Issue #701, opened 2015-09-04
Better support for database routing


foxx:
As an extension from the discussions in #582 and #583, I'd like to propose a few changes. The implementation of #583 was rejected on the basis of design, as it would have required a connection manager, something which `peewee` does not handle by design (and rightly so).

Playhouse module [ReadSlave][1] implements routing by overloading `select()` and injecting the correct database as determined from `_get_read_database()`, which allows for basic routing of read queries.

However, implementing write queries is significantly harder because methods such as `create_or_get()` reference `cls._meta` directly. I've been able to work around this by extending `ReadSlaveModel` with a custom metaclass, which I've included in this post for future ref, but it's a dirty hack at best.

As such, I would like to propose an extension to `Model`;

* Add two new methods on `Model` called `get_read_db()` and `get_write_db()`. Both of which return  `cls._meta.database` by default, ensuring backwards compatibility
* Modify anywhere which uses `cls._meta.database` to use these methods instead (e.g. `select()` uses `get_read_db()` and `create_or_get()` uses `get_write_db()` etc).

I've looked over the source and can't see any problems with implementing the above. It will ensure backwards compatibility whilst allowing developers to implemented true read/write routing.

I would be happy to contribute time putting together a PR for this.

Thoughts?

```python

class MetaOptionsMixin(object):
    """Late mix-in to provide simple database routing"""
    @property
    def database(self):
        """Return write db here"""
        return your.custom.db.stuff.here

class BaseModelMetaClass(peewee.BaseModel):
    """Add support for database routing"""

    # XXX: add proper database router support
    def __new__(cls, name, bases, attrs):
        """Add support for automatically registering models with app"""
        new_cls = super().__new__(cls, name, bases, attrs)
        new_cls._meta.db_name = new_cls._meta.database
        extend_instance(new_cls._meta, MetaOptionsMixin)
        return new_cls
    def selct

class Model(ReadSlaveModel, metaclass=BaseModelMetaClass):
    pass
```

[1]: https://github.com/coleifer/peewee/blob/master/playhouse/read_slave.py

stas:
Relevant https://github.com/coleifer/peewee/issues/689#issuecomment-133063492

coleifer:
> * Add two new methods on Model called get_read_db() and get_write_db(). Both of which return cls._meta.database by default, ensuring backwards compatibility
[...]

PR #800, opened 2015-12-28
Patch proposal for better database routing - see #701

foxx:
After several different attempts, this seems to be the cleanest way forward. 

For this patch to be merge worthy, I'd need sign off from @coleifer on these questions about `Using()`. Also feedback from @stas and @alexanderad would be appreciated

This has minimal code impact as `_execute()` seems to handle every edge case, whilst ensuring that queries are still executed in the correct database order (to the extent of all tests passing, there /might/ be untested edge cases).

On the model there is now `determine_database`, by default uses the same SQL string inspection method as `read slave`, but in the event that users are making heavy use of raw queries with `IF()` conditionals, then they can extend to their hearts content.

There is also now `get_read_database()` and `get_write_database()`, by default these continue to use `self._meta.database` but can be modified to whatever the user wants.

The remaining problems now rests with `Using()`, as it does not support two database args. Altering the syntax with extra keywords would make it backwards incompatible, therefore I'd propose that we support the following syntax;

```
Using(database)
Using((read_database, write_database))
```

The last gotcha is whether `Using()` should override model, or visa versa. In my personal opinion, if someone has specified a `Using()` context then this should override whatever is in `get_read_database()` and `get_write_database()`.

From what I can tell, this /should/ cover all use cases from #701.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2580,7 +2580,8 @@ class Query(Node):
 
     def _execute(self):
         sql, params = self.sql()
-        return self.database.execute_sql(sql, params, self.require_commit)
+        db = self.model_class.determine_database(self, sql, params)
+        return db.execute_sql(sql, params, self.require_commit)
 
     def execute(self):
         raise NotImplementedError
@@ -4399,6 +4400,21 @@ class Model(with_metaclass(BaseModel)):
 
         for k, v in kwargs.items():
             setattr(self, k, v)
+
+    @classmethod
+    def get_read_database(self):
+        return self._meta.database
+
+    @classmethod
+    def get_write_database(self):
+        return self._meta.database
+
+    @classmethod
+    def determine_database(self, query, sql, params):
+        if sql.lower().startswith('select'):
+            return self.get_read_database()
+        else:
+            return self.get_write_database()
 
     @classmethod
     def alias(cls):
--- a/playhouse/read_slave.py
+++ b/playhouse/read_slave.py
@@ -25,22 +25,9 @@ from peewee import *
 
 class ReadSlaveModel(Model):
     @classmethod
-    def _get_read_database(cls):
+    def get_read_database(cls):
         if not getattr(cls._meta, 'read_slaves', None):
             return cls._meta.database
         current_idx = getattr(cls, '_read_slave_idx', -1)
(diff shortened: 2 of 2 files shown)


peewee.py
```diff
@@ -4401,6 +4402,21 @@ def __init__(self, *args, **kwargs):
...
+    @classmethod
+    def get_write_database(self):
+        return self._meta.database
+
+    @classmethod
+    def determine_database(self, query, sql, params):
+        if sql.lower().startswith('select'):
+            return self.get_read_database()
+        else:
+            return self.get_write_database()
```
coleifer: I am concerned about negative performance implications with this code. Perhaps a better option would be to move this type of logic into the `read_slave` module. Although honestly, the best solution IMO would be a separate third party package.

coleifer:
Although I agree this is a pretty clean patch, I am going to pass on merging.

I think that database routing is a complex enough topic that I do not want to include it in peewee itself (as I don't want to maintain it). One option would be to create a new package `peewee-routing` or something similar that encapsulates this logic.

foxx:
Thanks for the quick reply! If I put myself forward as a maintainer for this, and submitted a fully documented PR, would you consider merging into playhouse? As a last option I could create a new repo, but would argue the use cases are common enough to justify being in the playhouse.

coleifer:
I appreciate the offer and don't want to seem to be discouraging, but I'm actively trying to rein in some of the sprawl in the code-base and tighten things up. Database-routing, to me, both complex and uninteresting from a maintenance perspective. It is my strong preference to keep it out of this repo.

foxx:
Okay no worries, I'll look into creating a separate repo for this. I'll update the original issue with the URL once finished, if you're happy with it then (and agree with the approach used), then perhaps it might be worthy of a small section in the docs, so at least people know where to look?

coleifer:
> then perhaps it might be worthy of a small section in the docs, so at least people know where to look?

I'm sure it'll turn out well, and I'd be happy to link it up.

peewee.py
foxx: I'll add some performance testing as part of the new repo and will include these numbers in the README, hopefully it will be minimal

PR #819, opened 2016-01-14
Fixed broken BlobField/BinaryField in MySQL

### LINKED ISSUE #818: BinaryField/BlobField/UUIDField is broken with MySQL
Looks like both `BlobField` and `BinaryField` in MySQL are actually broken, because MySQL expects a `max_length`, and thus all values are being truncated down to 1 byte.

The most obvious fix seems to be;

```
class BinaryField(BlobField):
    db_field = 'varbinary'

class FixedBinaryField(CharField):
    db_field = 'binary'    
```

The test should then probably be updated to include a `data` value comparison, and then add the appropriate database mappings after.

Submitting PR shortly

foxx:
Fixes #818

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -43,6 +43,7 @@ __all__ = [
     'BigIntegerField',
     'BlobField',
     'BinaryField',
+    'FixedBinaryField',
     'BooleanField',
     'CharField',
     'Check',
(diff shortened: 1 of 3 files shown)


foxx:
This will need sign off and review from @coleifer before merge. I'm about 90% happy with the code, but there could be some edge cases I haven't considered.

coleifer:
I'm leaning towards killing off the binary field altogether.

foxx:
@coleifer Yeah I just updated the original issue with my conclusion, lemme know your thoughts

foxx:
Closing as per #818

PR #821, opened 2016-01-15
Fallback gracefully in the face of ancient (and probably broken) Cython versions

jherre:
Some folks have dependencies that won't support easy Cython upgrades. Ignore incompatible Cython installations - same thing as not having it at all...

Tested on py27, py34 and py35 and all versions of Cython up to 0.23.4


coleifer:
Where did you arrive at that version number? I'm leery of merging this as I think it might cause some false positives.

jherre:
I ran a matrix of all versions of Cython since 0.21 using py27, py34 and py35 using the peewee tests.

Using versions prior to 0.22.1 doesn't work.  Failures happen during cythonization.  If it's able to install, all tests pass.   Trying to install peewee with any version prior to Cython==0.21.1 will fail.

coleifer:
What about py26? Peewee officially supports python 2.6.

jherre:
Yep, same result on python26.  Using versions of Cython earlier than 0.21.1 will cause setup to fail.

It's really about what versions of Cython you're will to support for speedups.  I think it's fair to say that folks need to upgrade to use the speedups in peewee but it's not fair to say they have to upgrade just to use peewee at all...

FWIW, Cython 0.22.1 was released 2014-10-18 which is really not that long ago in Python2.6 terms.

Issue #783, opened 2015-12-04
RetryOperationalError mixin failing on retry w/ error 2006


bakemecookies:
Tried using the new RetryOperationalError mixin and although it's reaching the exception block properly (known because it's going to shortcuts.py, line 208, execute_sql), the retry fails and raises the same error.

My hunch is that since we're using PooledMySQLDatabase, could it be that we're closing one stale connection but just getting another stale one on the retry?  Not that it's necessarily a solution, but in the past we've used a combo of sql_error_handler to close stale connections and recursion on execute_sql (i.e., infinite retry only in the case of 2006 errors, as opposed to the single retry in RetryOperationalError) within its exception block to handle this.

coleifer:
Unfortunately, I don't think I will be going back to the "sql error handler" API. I think the implementation in the `RetryOperationalError` mixin is an appropriate way to try to handle re-connect/re-try logic. Maybe you can make some changes to your pool implementation, particularly overriding the `_is_closed()` method to perform some sort of test to check that the conn is really active. Take a look at the pool module's code, particularly the `PooledMySQLDatabase._is_closed()` method...I bet you can find something that'll work for you.

PR #822, opened 2016-01-15
_is_closed check doesn't work with PooledMySQLDatabase and pymysql

decoratorz:
We encountered the same problem as described here: https://github.com/coleifer/peewee/issues/783

Our connections in the pool were closed by MySQL with the 2006 error after reaching server-side timeout. Because we don't always control the environment we run in, we didn't want to guess the my.cnf wait_timeout value and hardcode it in the `stale_timeout` argument for Pool, we just always need to restart a dead connection.

But the solution was simple. The main problem lies [here](https://github.com/coleifer/peewee/blob/7857f08a6fb70fc91c8e884d9eaa2289155d29f5/playhouse/pool.py#L168). This check doesn't work, because `pymysql.connection` also has the `open` attribute (at least in our version 0.6.6). We could use some other method to differentiate between  pymysql and MySQLdb, but there is actually no need. Even though `conn.ping` always [returns None](https://github.com/farcepest/MySQLdb1/blob/d236bb7f6b45f81f0156c4a0cfdef2319ba825d9/_mysql.c#L2011) on MySQLdb, it still does the right thing: throws exception on a dead connection.

After this fix PooledMySQLDatabase recognizes closed connections and everything works good.

### DIFF
--- a/playhouse/pool.py
+++ b/playhouse/pool.py
@@ -165,17 +165,10 @@ class PooledMySQLDatabase(PooledDatabase, MySQLDatabase):
     def _is_closed(self, key, conn):
         is_closed = super(PooledMySQLDatabase, self)._is_closed(key, conn)
         if not is_closed:
-            if hasattr(conn, 'open'):
-                # MySQLdb `ping()` seems to always return `None` in my testing.
-                # So the `open` attribute will be used instead.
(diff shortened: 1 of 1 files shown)


coleifer:
Nice, I tested this out and it appears to work correctly. Thanks!

PR #823, opened 2016-01-16
PaginatedQuery should not throw 404 on empty query

econne01:
Working through the [Flask Blog in an hour or less tutorial](https://github.com/econne01/peewee.git), I noticed 404 errors are being thrown when the database is empty.

I think `PaginatedQuery` class should gracefully handle an empty query and return empty set when the query returns no results.


coleifer:
The behavior is intended

econne01:
I see. Ok, thanks

Issue #836, opened 2016-01-28
apsw and non integer PK


ronyb29:
the apsw driver currently returns the last [rowid](https://www.sqlite.org/c3ref/last_insert_rowid.html) instead of the las inserted PK, this causes problems for any non int pk (like a UUID)

coleifer:
https://github.com/coleifer/peewee/pull/837#issuecomment-175938671

coleifer:
I see the issue now. Normally, last_insert_id() checks first that the model has an auto incrementing pk.

```python
def last_insert_id(self, cursor, model):
        if model._meta.auto_increment:
            return cursor.lastrowid
```

This check is missing in the apsw driver, though. Apologies for the confusion on my part. The fix will just need to add this check to the apsw implementation.

PR #837, opened 2016-01-28
fix for issue #836

ronyb29:
fix for issue #836

### DIFF
--- a/playhouse/apsw_ext.py
+++ b/playhouse/apsw_ext.py
@@ -103,7 +103,8 @@ class APSWDatabase(SqliteExtDatabase):
         return cursor
 
     def last_insert_id(self, cursor, model):
-        return cursor.getconnection().last_insert_rowid()
+        last_rowid = cursor.getconnection().last_insert_rowid()
+        return model.select().where(SQL('_ROWID_=?',last_rowid)).get()._get_pk_value()
 
     def rows_affected(self, cursor):
         return cursor.getconnection().changes()
--- a/playhouse/tests/test_apsw.py
+++ b/playhouse/tests/test_apsw.py
@@ -3,7 +3,7 @@ import datetime
 
 from playhouse.apsw_ext import *
 from playhouse.tests.base import ModelTestCase
-
+from uuid import uuid1
 
 db = APSWDatabase(':memory:')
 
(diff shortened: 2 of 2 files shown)


coleifer:
This is not the intended behavior of last_insert_id() -- the intended behavior is to return the last inserted ``rowid``. For auto-incrementing primary key fields, this works as expected.

If you have a user-defined primary key, then presumably you know what it is beforehand so there's no need to return it (and issue an extra query into the bargain).

ronyb29:
the thing is, currently, Model.save() sets the pk to whatever the database reports as `last_insert_id`, so when you make relationships with instances created in code (not retrieved from the DB) the foreign key for ForeignKeyField are broken.

coleifer:
Peewee only automatically sets the primary key for auto incremented integers. If you have a uuid primary key you presumably set that in the application. I'm confused as to what issue you're actually experiencing.

ronyb29:
Perhaps I have failed to convey the issue. Currently it is setting the PK regardless of its kind to whatever `last_insert_id ` returns. Please take a look at this tests.
 
```python
from uuid import uuid1, UUID

from playhouse.apsw_ext import *
from playhouse.tests.base import ModelTestCase

db = APSWDatabase(':memory:')

class BaseModel(Model):
    class Meta:
        database = db

class Batch(BaseModel):
    id = UUIDField(primary_key=True, default=uuid1)

class Document(BaseModel):
    id = UUIDField(primary_key=True, default=uuid1)
    batch = ForeignKeyField(Batch, related_name='documents', to_field='id')

class APSWPKTest(ModelTestCase):
    requires = [Batch, Document]

    def test_non_int_pk_saving(self):
        batch = Batch()
        self.assertIsInstance(batch.id, UUID)  # passes
        batch.save(force_insert=True) # Will save the right id in the DB, but will modify the instance and set it to the rowid
        self.assertIsInstance(batch.id, UUID)  # fails, contains the ROWID returned by last_insert_id

    def test_non_int_pk_relationships(self):
        batch = Batch.create()
        self.assertIsInstance(batch.id, UUID)  # fails, contains the ROWID returned by last_insert_id

        doc = Document(batch=batch)
        self.assertIsInstance(doc.id, UUID)  # passes
        doc.save(force_insert=True)
        self.assertIsInstance(doc.id, UUID)  # fails, contains the ROWID returned by last_insert_id
```


That is because Model.save actually sets the instance pk field after inserting
```python
class Model(with_metaclass(BaseModel)):
    ...
    def save(self, force_insert=False, only=None):
        ...
        self._populate_unsaved_relations(field_dict)
        if pk_value is not None and not force_insert:
            ...
        elif pk_field is None:
            ...
        else:
            pk_from_cursor = self.insert(**field_dict).execute()
            if pk_from_cursor is not None:
                pk_value = pk_from_cursor
            self._set_pk_value(pk_value)
            rows = 1
        self._dirty.clear()
        return rows
```

Perhaps a better solution than what I originally proposed would be for Model.save to only set the PK if it was an instance of PrimaryKeyField

PR #844, opened 2016-02-05
Allow to make dynamic related name with current class

yohanboniface:
When declaring a ForeignKey in an abstract class, this allows
to customize the related_name for each child.

For example:
```python
class Other:
    pass

class Abstract:
    # Will create Other.child1_foo and Other.child2_foo accessors.
    other = ForeignKey(Other, related_name='{classname}_foo')

class Child1(Abstract):
    pass

class Child2(Abstract):
    pass
```

Thanks :)

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1274,7 +1274,8 @@ class ForeignKeyField(IntegerField):
         return ReverseRelationDescriptor(self)
 
     def _get_related_name(self):
-        return self._related_name or ('%s_set' % self.model_class._meta.name)
+        return (self._related_name or '{classname}_set').format(
+                                        classname=self.model_class._meta.name)
 
     def add_to_class(self, model_class, name):
         if isinstance(self.rel_model, Proxy):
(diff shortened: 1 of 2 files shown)


coleifer:
What you have there is really no different from the current behavior. For instance:

```python
In [1]: from peewee import *

In [2]: class User(Model):
   ...:     username = CharField()
   ...:     

In [3]: class BaseTweet(Model):
   ...:     user = ForeignKeyField(User)
   ...:     

In [4]: class Tweet(BaseTweet):
   ...:     pass
   ...: 

In [5]: class PhotoTweet(BaseTweet):
   ...:     pass
   ...: 
In [6]: User.phototweet_set
Out[6]: <peewee.ReverseRelationDescriptor at 0x7f42ac136710>

In [7]: User.tweet_set
Out[7]: <peewee.ReverseRelationDescriptor at 0x7f42ac136490>
```

yohanboniface:
> What you have there is really no different from the current behavior.

Well, I know that. But I want to take control over this naming…

yohanboniface:
@coleifer can you reconsider this? In my point of view it does not hurt, while allowing user to customize easily related names.

Issue #845, opened 2016-02-05
APSWDatabase not respecting pragmas keyword.


cledoux:
When initializing an apsw database with pragmas, these pragmas seem completly ignored. This problem is not seen in the parent class, SqliteExtDatabase.

```
>>> db = APSWDatabase('/tmp/apsw.sqlite', pragmas=(('journal_mode', 'WAL'),))
>>> db.connect()
>>> db.execute_sql('PRAGMA journal_mode;').fetchone()[0]
'delete'
>>> db = SqliteExtDatabase('/tmp/sqliteext.sqlite', pragmas=(('journal_mode', 'WAL'),))
>>> db.connect()
>>> db.execute_sql('PRAGMA journal_mode;').fetchone()[0]
'wal'
```

cledoux:
I've created a pull request with test cases that illustrate the issue along with a potential fix.

APSWDatabase overwrites _connect without calling _add_conn_hooks. I added this missing call.

cledoux:
Thanks for the amazingly fast turnaround!

PR #846, opened 2016-02-05
Proposed fix for #845

cledoux:
APSWDatabase _connect() without also making a call to _add_conn_hooks().

This was resulting in the pragmas keyword for __init__ being ignored as the call to _set_pragmas  was never made.

I've added test cases that illustrate my original problem and added the call to _add_conn_hooks() to APSWDatabase._connect().


coleifer:
Merged a slightly different fix. Thanks for bringing this to my attention.

Issue #838, opened 2016-01-29
Custom ManyToManyQuery?


yohanboniface:
I'd like to implement a custom result wrapper for my ManyToManyQuery (already done a custom [SelectQuery](https://github.com/etalab/ban/blob/master/ban/db/model.py#L6)), but currently the `ManyToManyQuery` is hardcoded in the [ManyToManyFieldDescriptor](https://github.com/coleifer/peewee/blob/master/playhouse/fields.py#L168) (and in `ManyToManyQuery.clone`), which itself is hardcoded in [ManyToManyField](https://github.com/coleifer/peewee/blob/master/playhouse/fields.py#L116).

Would you consider a PR that would make that more customizable? I'm thinking about making those class attributes. Something like this:

```python
# Or maybe better at Field level directly?
class ManyToManyField(Field):
    descriptor_class = ManyToManyFieldDescriptor

    def add_to_class(self, model_class, name):
        …
        setattr(model_class, name, self.descriptor_class(self))
        …

class ManyToManyFieldDescriptor(FieldDescriptor):
    query_class = ManyToManyQuery
    …
```

Thoughts?

coleifer:
See f1988db for the changes.

yohanboniface:
Thanks for the quick fix @coleifer!

That fixes the main issues, though I'd still need to overwrite and copy-paste [`ManyToManyFieldDescriptor.__get__`](https://github.com/coleifer/peewee/blob/master/playhouse/fields.py#L169) and [`ManyToManyQuery.clone`](https://github.com/coleifer/peewee/blob/master/playhouse/fields.py#L189), any idea to prevent that too?

yohanboniface:
Replacing:
```
    def clone(self):
        query = ManyToManyQuery(
            self._instance,
            self._field_descriptor,
            self.model_class)
        query.database = self.database
        return self._clone_attributes(query)
```
by
```
    def clone(self):
        query = self.__class__(
            self._instance,
            self._field_descriptor,
            self.model_class)
        query.database = self.database
        return self._clone_attributes(query)
```
would help without hurting, imho :)

coleifer:
You could probably override `clone()` in your subclass. I think that would
actually be more appropriate.

yohanboniface:
> You could probably override `clone()` in your subclass. I think that would actually be more appropriate.

Why is that more appropriate?
[...]

PR #848, opened 2016-02-09
Use current class as reverse ManyToManyField (cf #838)

yohanboniface:
Thanks :)


yohanboniface:
Humm, why? :s
That does not seem to hurt anything while allowing more DRYer overriding of ManyToManyField.

Issue #858, opened 2016-02-17
Add support for additional connect args as query string in db_url


RealSalmon:
I'm happy to take a stab at this, but I thought it best to first propose it prior to spending any time . . . 

When using playhouse.db_url for connections, it is not currently possible to include additional connect args directly in the connection string. These must instead be supplied as separate keyword args.

```python
connect('mysql+pool://user:passwd@ip:port/my_db', max_connections=20, stale_timeout=300)
```

It would be nice to get those extra connection arguments directly from the connection string itself

```python
# DATABASE_URL has mysql+pool://user:passwd@ip:port/my_db?max_connections=20&stale_timeout=300
connect(os.environ['DATABASE_URL'])
```

Thoughts?

coleifer:
Yes!

RealSalmon:
Great, I'll start messing around with this evening (real life permitting . . . ).

tomako:
That's cool but how could we set False (eg. register_hstore) via url? Should parse the query string with `keep_blank_values=True`, right?

PR #859, opened 2016-02-18
Get extra connection args from query string in db_url

RealSalmon:
### Changes
- As discussed in #858, this adds support for supplying extra connection args in the query string when using playhouse.db_url

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -1,7 +1,7 @@
 try:
-    from urlparse import urlparse
+    from urlparse import urlparse, parse_qsl
 except ImportError:
-    from urllib.parse import urlparse
+    from urllib.parse import urlparse, parse_qsl
 
 from peewee import *
 from playhouse.pool import PooledMySQLDatabase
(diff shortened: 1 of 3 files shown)


playhouse/db_url.py
```diff
@@ -42,6 +42,11 @@ def parseresult_to_dict(parsed):
     elif 'sqlite' in parsed.scheme and not connect_kwargs['database']:
         connect_kwargs['database'] = ':memory:'
 
+    # Get additional connection args from the query string
+    connect_kwargs.update(
+        {k: v[0] for k, v in parse_qs(parsed.query).iteritems()}
+    )
+
```
coleifer: Should change to use `parse_qsl`

coleifer:
Looks good, just one thing to change and I'll be happy to merge.

RealSalmon:
Actually that dictionary comprehension breaks in Python 2.6 (forgot about that). I didn't notice it before because travis didn't pick up the PR initially.

RealSalmon:
A few fixes were required for Python 2.6.

playhouse/db_url.py
```diff
@@ -42,6 +51,11 @@ def parseresult_to_dict(parsed):
     elif 'sqlite' in parsed.scheme and not connect_kwargs['database']:
         connect_kwargs['database'] = ':memory:'
 
+    # Get additional connection args from the query string
+    connect_kwargs.update(
```
coleifer: Can you not just do `connect_kwargs.update(parse_qsl(query))`? I think dict.update takes a tuple-of-2-tuples.

RealSalmon: Agreed. I suffered from tunnel vision here, caused by the late hour and having previously used a dict comprehension to parse the results of parse_qs . . . followed by the Python 2.6 problems.

PR #904, opened 2016-04-02
Removed side effects while cleaning prefetch query

kammala:
There was a strange behaviour: then you do prefetch on complex query with some grouping, you will have some kind of exception or incorrect result. The reason of this is simple: `clean_prefetch_query` removes any information about grouping from query before placing it as subquery in prefetch queries, but it is done on original query instead of cloned one. 

This pr fixes this little bug(in a single code line ^_^).


coleifer:
Nice pull request, and nice catch. Thanks!

PR #905, opened 2016-04-05
doc shows wrong default value for wrapped_count parameter

lixxu:
`wrapped_count` has default value `False` for `clear_limit`, but the doc shows `True`


coleifer:
Thank you for this!

PR #911, opened 2016-04-12
Change description of options in Using class docs

ahessert:
Previous description implied that the default for the "with_transaction" kwarg was False, while the default is True. Developers must specify "with_transaction=False" if they don't want to use a transaction block with the Using class context.


coleifer:
Thanks!

PR #944, opened 2016-05-23
fix the support of schema for PostgreSQL when we try to create a new table

matrixise:
In the Meta data of a class, we can specify the `schema` attribute, but this one is not used when we try to create the table. in fact the schema must be existing. in this case, we will create it after the creation of the table.

I prefer to propose a solution, I can rewrite it if necessary.


matrixise:
ok, my branch has some issues with travis, I will try to fix it asap

PR #949, opened 2016-05-31
make genned SQL easier on the eyes by removing unnecessary quotes (ex…

keredson:
…: `select id from posts` instead of `select "id" from "posts"`)

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1586,6 +1586,7 @@ class QueryCompiler(object):
         }
 
     def quote(self, s):
+        if re.match(r'[a-zA-Z]\w*$', s): return s
         return '%s%s%s' % (self.quote_char, s, self.quote_char)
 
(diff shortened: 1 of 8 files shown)


coleifer:
Just noting for posterity, this is a horrible suggestion because it requires:

* The maintenance of a cross-vendor list of all SQL keywords to avoid invalid SQL (i.e. only quote names if they correspond to a SQL keyword)
* Inspecting each named entity to see if it contains spaces or other characters that would otherwise require quoting

All for ... fewer quotes in the SQL that this library is designed to abstract away. The idea is not a bad one in the sense that, yes, fewer quotes would make the sql more easy to read. But it is a bad one in the way it forces the implementation to become special-cased in numerous ways. Special cases are bad and are avoided as much as possible in peewee. It was for this latter reason, more than anything else, that this was closed without merge.

PR #952, opened 2016-06-01
make genned table aliases more human-friendly

keredson:
instead of: `select t1.id from ledger_entry as t1`
generate the alias from the table name: `select le1.id from ledger_entry as le1`

where the alias is an acronym of the table name.  makes it easier to read large joins w/ many tables.

@coleifer unit tests not updated yet.  i want to know if you're ok-with/opposed to this before i spend the time on that part.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -27,6 +27,7 @@ import threading
 import uuid
 from bisect import bisect_left
 from bisect import bisect_right
+from collections import defaultdict
 from collections import deque
 from collections import namedtuple
 try:
@@ -1400,7 +1401,7 @@ class AliasMap(object):
 
     def __init__(self, start=0):
         self._alias_map = {}
-        self._counter = start
+        self._counters = defaultdict(lambda: start)
 
     def __repr__(self):
         return '<AliasMap: %s>' % self._alias_map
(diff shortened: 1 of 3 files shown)


coleifer:
I'm going to pass on this one as well. Seems like a special-case that really isn't that special. If you want to see an example of a custom query compiler, check out the tests -- there is one that will use the table name instead of `t{\d}`.

keredson:
this is just totally bizarre to me.  why would you **prefer** meaningless table alias names?  what possible benefit is that?

coleifer:
If you're concerned about table aliases, you can always specify a default to use with a given model. Again, this is [documented](http://docs.peewee-orm.com/en/latest/peewee/models.html?highlight=table_alias#model-options-and-table-metadata).

```python
class MyModel(Model):
    class Meta:
        table_alias = 'mm'
```

And @keredson tell your little buddies their emojis are really not helping anybody.

coleifer:
> this is just totally bizarre to me. why would you prefer meaningless table alias names? what possible benefit is that?

Can you please dial down your outrage? Can you dial down your tone? It's not helpful.

coleifer:
For posterity, this is not a bad idea in that using the table's name to create the alias will make the SQL more easy to scan. It was for the following reasons I decided not to merge this:

* Requires the addition of code to "fix" something that is technically not broken.
* The purpose of Peewee is to abstract away the process of writing SQL.
* You can explicitly set a model's table alias per-query or globally using existing APIs.
* The implementation is (subjectively) gross.
* Requires rewriting a huge number of test assertions.

Another misguided PR, not aligned with peewee's goals or aesthetic.

PR #957, opened 2016-06-02
refactor _joins to use a defaultdict

keredson:


keredson:
hah!  these unit tests are infuriating.  not sure how to reproduce this locally.  going to try some test commits...

coleifer:
How is this any better? I'm going to pass.

keredson:
it's simpler and eliminates a class of bugs where one might forget to do the "if (does this key exist?) append else set" conditional check.
https://docs.python.org/2/library/collections.html#defaultdict-objects

Issue #959, opened 2016-06-03
first() should set a `LIMIT 1`


keredson:
when looking at https://github.com/coleifer/peewee/issues/958, i noticed:

```python
class Person(pw.Model):
  name = pw.CharField()
  class Meta:
    database = db

alice = Person.create(name='Alice')
print Person.select().where(Person.name=='Alice').first()
```

generates this sql:

```sql
SELECT "t1"."id", "t1"."name" FROM "person" AS t1 WHERE ("t1"."name" = ?)
```

why would this not set a `LIMIT 1`?  even if the library code stops the underlying sql lib's iteration it's likely already transferred way more data than that.  (psql's default is 1000 rows at a time IIRC)

coleifer:
It is a design decision...and documented. Seriously what is wrong with you.

http://docs.peewee-orm.com/en/latest/peewee/api.html#SelectQuery.first

keredson:
> Seriously what is wrong with you.

i don't see the utility of your personal attacks here.

maybe it was a design decision, maybe it was a post-hoc justification of an unintentional implementation quirk, it's impossible for me to tell.  either way "caching the entire result set when the user asked for the first record because they *might* want more later" is a weak argument for such surprising behavior.  is questioning that choice (regardless of how you got there) not legitimate in your mind?

vkhvst:
@keredson makes a reasonable argument for setting `limit 1` especially because the result set can be very large. Maybe the caching can be moved to a different method where that behavior is expected or more intuitive

arikfr:
To be honest, I find this behavior surprising too. It's true the documentation mentions this, but maybe it's worth putting a clearer message about this. @coleifer would you mind a pull request with this kind of update to the documentation?

keredson:
just to be clear, i'm not advocating changing this behavior in a point release or anything.  but if a major 3.0 rewrite is in the works (which @coleifer has already indicated would be API changing), that seems like a reasonable opportunity.  (*possibly* in v2.9 if there was a clear consensus, or some version of `from peewee.__future__ import first`, or any number of other obvious options, but all that discussion seems secondary to "is this a good idea at all?")

perhaps a better name for the existing behavior would be `peek()`?

coleifer:
Context of my earlier comment... https://github.com/coleifer/peewee/issues/958#issuecomment-223492592

At any rate, I think the documentation should be more clear and would accept a PR or can just type it up real quick either way.

[...]

PR #960, opened 2016-06-03
First limit 1

keredson:
for possible future inclusion (not 2.8.2) regarding https://github.com/coleifer/peewee/issues/959 or if anyone else wants to pull it into their forks.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2917,6 +2917,12 @@ class SelectQuery(Query):
                 % self.sql())
 
     def first(self):
+        try:
+            return list(self.limit(1))[0]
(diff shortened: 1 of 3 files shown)


coleifer:
I'd like to preserve the behavior where multiple calls to `first()` does not yield multiple queries. I *do* agree, though, that there is benefit in applying a limit and that a second `peek()` method might more correctly. I've gone ahead and implemented this in a separate patch. You can compare notes to see how I've done this a bit differently: eeb6d4d727da8536906a00c490f94352465e90bb

keredson:
+1

Issue #958, opened 2016-06-03
get() should throw an error if more than one row exists


keredson:
```python
import peewee as pw

class Person(pw.Model):
  name = pw.CharField()
  class Meta:
    database = db

alice = Person.create(name='Alice')
alice2 = Person.create(name='Alice')

print Person.get(Person.name=='Alice')
```

this generates the following SQL:
```sql
SELECT "t1"."id", "t1"."name" FROM "person" AS t1
WHERE ("t1"."name" = ?) LIMIT 1 OFFSET 0
```

if the conditional criteria don't uniquely identify a record, you're effectively randomly choosing which row is returned to the user when the user has clearly indicated a belief that the record is unique.  as is this method is simply a shorthand for `.first()`.

it's fairly common for this feature to be implemented with a `LIMIT 2`, where the lib will throw some sort of `NotUniqueException` if more than one row are returned.

i'll be happy to implement if you don't disagree with it on principal.

coleifer:
Sorry but that has always been a design decision I've liked, and I'm not going to change it.

coleifer:
This is documented anyways, did you even bother to read it?

http://docs.peewee-orm.com/en/latest/peewee/api.html#Model.get

keredson:
just because the quirkiness of the API is documented doesn't mean there isn't room for improvement.  and i've seen no documentation on *why* you like this this design decision over other alternatives.

coleifer:
Who do you think you are?

keredson:
> Who do you think you are?

a developer who generally likes your API, wants to use it and contribute back improvements.  is that not welcome here?

looking back at your accepted external PRs i see many regarding minor bug fixes and documentation tweaks, but nothing beyond that, nor any other non-trivial collaborators.  at the risk of being blunt:  is this project just your baby/territory and all worthwhile ideas flow only from your brain, and all questions regarding even the most trivial design decisions personally offensive to you?

coleifer:
I think you've got it backwards. You're new to the project, yet have submitted 5 issues and 4 pull-requests. Three of these issues were not actual issues - in two cases you had not read the docs, and in another (failing tests) you didn't see that it had been patched almost 2 weeks ago with a very clear commit message.  The other two (which includes this one), you state that you disagree with the semantics of methods **which are clearly documented**.

[...]

PR #962, opened 2016-06-03
Get should throw exception if not unique

keredson:
again, behavior changing, so not for any point release, but hopefully some future incantation.
re: https://github.com/coleifer/peewee/issues/958


coleifer:
I've commented already that the semantics of get() are not going to change, so I will pass on this change.

keredson:
3.0 man, 3.0!  :-)

coleifer:
It's not changing in 3.0 either man.

Issue #964, opened 2016-06-06
peewee should not gen SQL it knows is invalid


keredson:
```python
Post.select().where(Person.name=='derek').get()
```

generates:

```sql
SELECT "t1"."id", "t1"."title", "t1"."author_id" FROM "post" AS t1 
WHERE ("t2"."name" = ?) 
LIMIT 1 OFFSET 0
```

which is then executed by the database, which throws an error, as `"t2"."name"` doesn't exist.

(IMO) it should throw some sort of unreferenced table exception, rather than relying on the database to throw the error, since it already knows that it'll never work.

this code example is admittedly contrived (and obvious what the bug is), but for more complicated queries an error thrown earlier makes the API easier to use (as the error is more explicit about the root cause).  and the check is cheap.

i have a fix for this locally.  it's intertwined with some other stuff, but i'll be happy to pull it into a PR if it's agreed this is the correct behavior.

coleifer:
How would peewee do this in a performant, reliable manner?

Furthermore, if the error gets handled by the database, why is that so much worse than having it raised by Peewee?

keredson:
> How would peewee do this in a performant, reliable manner?

so in https://github.com/coleifer/peewee/blob/master/peewee.py#L1485 `AliasMap` offers a dictionary lookup which has a side effect of adding the object to the dict if it's not already there.  this is useful for generating the aliases in the first place in the from/join clauses, but the where clause should not be triggering new table aliases.  in my fix i added a second lookup method that didn't have the side effect, only called by `_parse_field`.

> Furthermore, if the error gets handled by the database, why is that so much worse than having it raised by Peewee?

i'm not writing the SQL, peewee is.  and peewee knows it's writing SQL that should never work.  and errors like `no such column: t2.name` are opaque.  in this simple example it's obvious, but outside of that it's not.  Something like `UnreferencedTableException: Table 'Person' isn't being selected from, therefore condition 'Person.name==X' isn't a valid constraint.` is more useful.

I'm not saying this is some major make or break thing here.  It's just a minor tweak to make it a little better.

coleifer:
> but the where clause should not be triggering new table aliases

Thats a pretty good idea, my concern is if it actually holds true in all cases. If you had a correlated subquery in the where clause, for instance? Not sure but there may be other edge cases?

keredson:
i think so, but it's possible i don't understand the full scope of what `_parse_field` does.  simple inner queries work at least:
```
[...]

PR #969, opened 2016-06-07
Throw error no alias

keredson:
for https://github.com/coleifer/peewee/issues/964.  i know you said no interest, but logging here anyway since i was doing it for my own fork regardless.

the `_join` wound up being unnecessary as far as i can tell.  i couldn't find a path in which it didn't hit the alias_map.

also added a unit test for your concat concern.

changing the unit test because it genned `t4` instead of `t3` i didn't like, but in the end the alias name used doesn't matter, and because if the way the counters are being passed around i'm not sure how fixable it is to keep the `t3`.  (and in my branch all that's gone anyway, so the personal ROI is low to continue debugging)

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1772,7 +1772,7 @@ class QueryCompiler(object):
             if query._from is None:
                 clauses.append(model.as_entity().alias(alias_map[model]))
             else:
-                [alias_map[x] for x in query._from] # seed the alias map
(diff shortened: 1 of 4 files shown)


coleifer:
Thanks for sharing the changes. Closing as wontfix per earlier discussion.

PR #971, opened 2016-06-07
insert_many: pass on `validate_fields` parameter

jberkel:


jberkel:
I came across this while looking at trying to work around another bug (insert_many + FTS5). Will follow up with another PR.

coleifer:
Awesome catch, thank you!

Issue #972, opened 2016-06-07
allow the user to specify the table alias name


keredson:
currently this:

```python
UA = User.alias()
UA.select().where(UA.username=='derek')
```

generates this:

```sql
SELECT "t1"."id", "t1"."username" FROM "users" AS t1
WHERE ("t1"."username" = ?)
```

we should allow the user to optionally specify the table alias.  example:

```python
UA = User.alias('u1')
UA.select().where(UA.username=='derek')
```

which would generate:

```sql
SELECT "u1"."id", "u1"."username" FROM "users" AS u1 
WHERE ("u1"."username" = ?)
```

optionally, a little syntax sugar that i prefer:
```python
UA = User.as_('u1')
UA.select().where(UA.username=='derek')
```

also identical aliases in the same query are handled appropriately, so aliases can be used in one-liners:

```python
User.as_('u1').select().where(User.as_('u1').username=='derek')
```

coleifer:
In rewriting the SQL generation, I've got this feature covered. The implementation of this in peewee 2.8 just feels a bit too hacky to me to want to merge.

PR #973, opened 2016-06-07
Impl custom alias names

keredson:
implements https://github.com/coleifer/peewee/issues/972


coleifer:
Oof...this implementation feels very fragile and hacky, and I don't think the extra lines of code are worth the feature of being able to specify an alias name.

Going to pass on this for now. The 3.0 stuff I'm working on can do this as a matter of course, so it'll land then.

keredson:
+1 on it for the rewrite.  including the alias equality?

PR #982, opened 2016-06-13
Prevent the silent overwriting of model's attributes e.g. "foo_id" when "foo" foreign key added

quantumdark:
In complex structures with a large number of fields easy to miss the collision, especially if it happened with parrent's model attribute.
Considering that the python dict doesn't guarantee the order of items, there are two possibilities:
1. The *_id attr will be overwritten by ObjectIdDescriptor and will no longer available, but field (if the attr is a field) still will be in _meta.fields. This leads to unexpected and floating bugs.
2. ObjectIdDescriptor will be overwritten by custom attribute. Much less harmful and may be unnoticed for a long time.

I think it should explicitly show the possible issue, by throwing an exception.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1400,6 +1400,15 @@ class ForeignKeyField(IntegerField):
                 raise AttributeError(error % (
                     self.model_class._meta.name, self.name, self.related_name))
 
+        if hasattr(model_class, name + '_id'):
+            name_id_attr = getattr(model_class, name + '_id')
+            if not (isinstance(name_id_attr, ForeignKeyField) and
+                    (name_id_attr.db_column == self.db_column)):
+                error = ('Foreign key: %s.%s id descriptor "%s" collision with'
+                         ' model attribute of the same name.')
+                raise AttributeError(error % (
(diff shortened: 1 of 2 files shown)


coleifer:
I'm kind of curious what kind of model structure you'd have to have to run into this kind of thing... I understand that there may be a bug here, but I'm not sure how you would actually encounter it in practice?

quantumdark:
I run into it not only once, last time we had save parameters from external API call in the model Inherited from the base model. In this case, the log of actions associated with a subscription entity.
Our deveveloper did'nt know the _id thing and just save it like that:

if v.validate(self.message):
    FortumoLog.create(
       subscription = self.subscription
       **self.message)

Nobody has noticed that one of the parameters is not stored (it just stored for history), but sometimes it drove to crashes our flask-admin, only on the prod with disabled debugging.) with some error 
"Failed to find field for filter: %s"
So I think it may save a few hours for someone.

coleifer:
Fixed 34e56cb887751453680641695bdec1fc6518b2cf

PR #988, opened 2016-06-15
Add mention in the doc about calling order_by() without argument to remove the ORDER BY clause.

yohanboniface:
Note: my first attempt was to call `order_by(None)`. Would you mind a PR to allow also this signature?


coleifer:
I've committed a fix for this, thanks for bringing it to my attention!

PR #1001, opened 2016-06-28
Set composite PK

kepkin:


coleifer:
What is the use-case for this?

PR #1003, opened 2016-07-03
add "is_not", "not_in" django-style filter support

qulc:
peewee filter not support `isnull`
https://docs.djangoproject.com/en/1.9/ref/models/querysets/#isnull

so i want add 'is_not' mapping to `OP.NOT_IN`
```python
Blog.filter(user__username__not_in=['u1', 'u2'], pub_date__is_not=None)
```


coleifer:
This is a nice PR, very clean, but the Django-style querying is deprecated and I do not plan on adding new functionality as it will be removed soon in the next major version...which hopefully isn't too far away!

PR #1004, opened 2016-07-05
Added a pip cache to Travis

thedrow:


coleifer:
What's going on in this PR? I'm not sure I understand the differences here.

thedrow:
It caches the compilation of the C extensions and also doesn't download pip dependencies.
The build as a result should be faster.

coleifer:
Just looking at the build times it doesn't appear to have made a difference.

coleifer:
Did a quick skim of the travis cache docs, I'm going to pass on this one.

PR #1009, opened 2016-07-11
Accept DATABASE_URL config variable in flask_utils

joeyespo:
Makes working with Heroku's `DATABASE_URL` standard slightly more straightforward, since Flask Utils already works with a connection URL out-of-the-box.


joeyespo:
Your implementation is better. :+1: 

Thank you :smiley:

PR #1016, opened 2016-07-19
The update of BooleanField coerce

mobishift2011:
Good afternoon, coleifer. We've found the most common case is that the data from the client(as browser) will send true or false, 1 or 0 to the server, but the conversion should be decided by peewee on the data field type. So may be the update of mine makes common sense.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1265,16 +1265,17 @@ class TimestampField(IntegerField):
 
 class BooleanField(Field):
     db_field = 'bool'
+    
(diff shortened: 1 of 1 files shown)


coleifer:
Is this in reference to a particular issue?

chenfuzhi:
**Amazing！**

coleifer:
This patch seems more suitable for a subclass of BooleanField that is designed to handle things like strings or untrusted data.

Issue #1018, opened 2016-07-19
Implement alternative DeferredRelation semantics


brenguyen711:
The [semantics behind circular foreign key dependencies](http://docs.peewee-orm.com/en/latest/peewee/models.html#circular-foreign-key-dependencies) get kind of unwieldy when you have more than a few models (and they're spread over multiple files).  This is because the `DeferredRelation` object has to be defined, used, then the other model defined in another file, then `set_model` has to be called on the original, and then you're left with the object reference dangling around that has no purpose.  IE the example in the docs:

```python
# Create a reference object to stand in for our as-yet-undefined Tweet model.
DeferredTweet = DeferredRelation()

class User(Model):
    username = CharField()
    # Tweet has not been defined yet so use the deferred reference.
    favorite_tweet = ForeignKeyField(DeferredTweet, null=True)

class Tweet(Model):
    message = TextField()
    user = ForeignKeyField(User, related_name='tweets')

# Now that Tweet is defined, we can initialize the reference.
DeferredTweet.set_model(Tweet)
```

It would be better if it all happened in the model definition with an optional parameter given to `DeferredRelation`.  Like:

```python
class User(Model):
    username = CharField()
    # Tweet has not been defined yet so use the deferred reference.
    favorite_tweet = ForeignKeyField(DeferredRelation('Tweet'), null=True)

class Tweet(Model):
    message = TextField()
    user = ForeignKeyField(User, related_name='tweets')
```

This would remove the need for the extra variable in the global namespace and the coordination of it over multiple files.  And since the parameter is optional, it can be fully backwards-compatible with the old syntax.

Thank you!

coleifer:
That's a very nice API. Depending on the class name (a string) is something I try to shy away from, as errors can occur despite the code appearing to be perfectly valid, for instance a typo or you happen to change the model name in one place and not the other.

I like the API, though, and agree that it would be worth supporting if done in a backwards-compatible way. Thanks for the patch, looking at it now.

coleifer:
Awesome. Thanks, I've merged your patch, really appreciate you taking the time to fix this one and implement such a clean patch.

PR #1019, opened 2016-07-19
impl new DeferredRelation semantics

brenguyen711:
this implements issue #1018

### DIFF
--- a/playhouse/tests/test_keys.py
+++ b/playhouse/tests/test_keys.py
@@ -334,16 +334,27 @@ class TestDeferredForeignKey(ModelTestCase):
         Language.drop_table(True)
         Language.create_table()
         Snippet.create_table()
+        Snippet2.drop_table(True)
+        Language2.drop_table(True)
+        Language2.create_table()
+        Snippet2.create_table()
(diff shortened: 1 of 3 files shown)


playhouse/tests/test_keys.py
```diff
@@ -334,17 +334,28 @@ def setUp(self):
...
 
     def test_field_definitions(self):
         self.assertEqual(Snippet._meta.fields['language'].rel_model, Language)
         self.assertEqual(Language._meta.fields['selected_snippet'].rel_model,
                          Snippet)
 
+    def test_field_definitions2(self):
+        self.assertEqual(Snippet2._meta.fields['language'].rel_model, Language2)
+        self.assertEqual(Language2._meta.fields['selected_snippet'].rel_model,
+                         Snippet2)
```
coleifer: Since the test-case is only testing the internal APIs, I think it'd be acceptable to move the model definitions into the test method.

I'll merge and then update the code.,

Thanks!

brenguyen711:
Awesome, thanks for the merge!

PR #1024, opened 2016-07-31
SqliteMigrator: fix edge case in rename_column

jberkel:
When renaming a column with the same name as the table
the table name will get accidentally replaced in the “CREATE index”
query:

```sql
CREATE INDEX
 "table_name_first_column_second_column" ON   
 "table_name" ("first_column", "second_column")
```


jberkel:
Actually there are still some cases where this might fail. Maybe there's a better way to recreate the indexes?

coleifer:
Fixed in 5aa587c

PR #1028, opened 2016-08-05
Pass kwargs from ensure_join to the join call.

paulbooth:
Allows join_type to be specified in ensure_join.

This will also allow any future arguments to join to automatically be handled through ensure_join.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2742,12 +2742,12 @@ class Query(Node):
         """Change or reset the query context."""
         self._query_ctx = model_class or self.model_class
 
-    def ensure_join(self, lm, rm, on=None):
+    def ensure_join(self, lm, rm, on=None, **join_kwargs):
         ctx = self._query_ctx
         for join in self._joins.get(lm, []):
             if join.dest == rm:
                 return self
-        return self.switch(lm).join(rm, on=on).switch(ctx)
+        return self.switch(lm).join(rm, on=on, **join_kwargs).switch(ctx)
 
     def convert_dict_to_node(self, qdict):
         accum = []


coleifer:
Looks pretty good, just hesitating on pressing "Merge" because it slightly changes the API in a backwards-incompatible way. Then I thought "Oh, well, the method is not documented, so nobody's really using it." And *then* I thought, "well, this guy's using it."

So, since I'm curious, what are you using this method for? It's a pretty obscure little helper function...

Secondly, what do you think about leaving `on=None` and just passing it along with the `**kwargs` to the `join()` method?

paulbooth:
We needed something like this for our Customer model with a nullable foreign key to Address. When we pass in a customer query to filter according to our filter language that includes Address, we need to make sure that Address has been joined without doing unnecessary switch and joins. So, we want to ensure_join on Address, but it is nullable and we want to JOIN_LEFT_OUTER, which we can't currently do.

Interestingly, we can actually get around this using the non-private method get_query_meta and inspecting the join_dict, but this is a bit of hack:
```
_select_fields, join_dict = customers.get_query_meta()
    for join in join_dict[Customer]:
        if join.dest is Address:
            break
    else:
        customers = customers.switch(Customer).join(Address, JOIN.LEFT_OUTER)
```

re: keeping on=None: Great idea, keep backwards-compatibility and future-proofing against any new join kwargs, sounds like a winner to me. Future code spelunkers might wonder why `on` is provided separately from other kwargs, but this should be obvious to allow for positional invocations. Change inbound

coleifer:
Awesome, thanks for this patch!

PR #1040, opened 2016-08-13
Add pool option to db_url.connect

joeyespo:
This separates the URL concern from the pooling concern. This is also backwards-compatible, so you can still use 'scheme+pool' if you want to hardcode the decision to use pooling right in your URL.

This lets you write:

```py
db = db_url.connect(app.config['DATABASE_URL'], pool=True)
```

Now, for example, Heroku's raw `DATABASE_URL` can be used as-is *and* have pooling enabled.

I had this logic in my app, but it made sense to me to expose it to `db_url`. Note that you can also ensure pooling is disabled by specifying `pool=False`, which I included for completeness.

If you think adding a new parameter is better (`def connect(url, pool=None, **connect_params):`), I'll change it to that. I just didn't want to break anybody. In the far future when Peewee drops Python 2 support, this can be accomplished with `def connect(url, *, pool=None, **connect_params):`.

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -1,7 +1,7 @@
 try:
-    from urlparse import urlparse, parse_qsl
+    from urlparse import urlparse, urlunparse, parse_qsl
 except ImportError:
-    from urllib.parse import urlparse, parse_qsl
+    from urllib.parse import urlparse, urlunparse, parse_qsl
 
 from peewee import *
 from playhouse.pool import PooledMySQLDatabase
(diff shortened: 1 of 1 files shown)


coleifer:
I will pass on this feature. The whole point of `db_url` is to allow you to specify things in the URL, and I don't see this as adding any particular value beyond what's already there.

joeyespo:
Ok.

Seems odd that when URL comes from an external source, you have to manually inject the Peewee-specific '+pool' argument. Oh well. Thanks for the response.

PR #1061, opened 2016-09-14
fixed a problem with TimestampField resolution

akrs:
When retrieving a value from the database, TimestampField didn't properly scale the microseconds by the resolution. Instead of using the resolution, it would always assume the fractional portion was in microseconds.

This fixes that by scaling the fractional second portion as a function of the resolution.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1277,7 +1277,9 @@ class TimestampField(IntegerField):
             if value == 0:
                 return
             elif self.resolution > 1:
-                value, microseconds = divmod(value, self.resolution)
+                ticks_to_microsecond = 1000000 // self.resolution
+                value, ticks = divmod(value, self.resolution)
+                microseconds = ticks * ticks_to_microsecond
                 return self._conv(value).replace(microsecond=microseconds)
             else:
                 return self._conv(value)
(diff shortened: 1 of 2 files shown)


akrs:
Also had to fix the unit test, as the unit test assumed that `654321` microseconds stored with a millisecond precision would come out as `654` microseconds instead of `654000` microseconds.

playhouse/tests/test_fields.py
```diff
@@ -453,7 +453,7 @@ def test_timestamp_field(self):
 
         expected = datetime.datetime(2016, 1, 3, 12, 12, 13)
         self.assertEqual(t3_db.local_us, expected.replace(microsecond=654321))
-        self.assertEqual(t3_db.utc_ms, expected.replace(microsecond=654))
+        self.assertEqual(t3_db.utc_ms, expected.replace(microsecond=654000))
```
coleifer: Ouch, my assertions were straight up testing the wrong value. Regret I didn't think it through a bit more, and apologies if this caused you all any problems.

peewee.py
```diff
@@ -1277,7 +1277,9 @@ def python_value(self, value):
             if value == 0:
                 return
             elif self.resolution > 1:
-                value, microseconds = divmod(value, self.resolution)
+                ticks_to_microsecond = 1000000 // self.resolution
+                value, ticks = divmod(value, self.resolution)
+                microseconds = ticks * ticks_to_microsecond
```
coleifer: Instead, I believe this should work:
```python
value, fraction = divmod(value, self.resolution)
microseconds = fraction * self.resolution
```

coleifer: Nevermind, I was using 3 as the resolution and this just happens to work, but the divide by 1000000 is necessary.

PR #1094, opened 2016-10-11
Easier way to define index type for Postgres custom fields

yohanboniface:
Thanks :)


coleifer:
Not sure what the purpose is for this change, can you explain?

yohanboniface:
> Not sure what the purpose is for this change, can you explain?

Allow to deal with `index_type` without the need of inheriting from `IndexedFieldMixin`.
The simpler, the better :)

coleifer:
I appreciate the simplicity of this... let me think about it. Thanks!

PR #1107, opened 2016-10-24
some string can't been decoded by utf8, return s

bagel:


coleifer:
Nice, thanks for this.

PR #1108, opened 2016-10-24
Added a test to check whether a connection is already open.

romanek-adam:
Most importantly the check prevents from leaking database connections,
which can be disasterous when using connection pools.

Moreover it allows one to catch errors in the code which were caused
by not using database.connect() / database.close() properly.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -3601,6 +3601,8 @@ class Database(object):
             if self.deferred:
                 raise Exception('Error, database not properly initialized '
                                 'before opening connection')
+            if not self._local.closed:
+                raise Exception('Connection misuse detected (already connected)')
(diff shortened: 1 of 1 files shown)


coleifer:
I'd like to add something like this, but am worried about breaking too many people's code... I'll look into a similar patch.

coleifer:
9e76c999868197a4911fa2fa457c50ab42431ff1

romanek-adam:
Thanks!

bnaul:
FYI: this did break our code like you mentioned above. I can work around it in our codebase but I expect this caused problems for a lot of people; is there any more robust way of doing this?

coleifer:
Just run `database.get_conn()` which will only open a connection if the DB is closed.

PR #1113, opened 2016-10-27
Add IntegrityError check in case post title is already in database.

bdauer:


bdauer:
Was just running through your awesome tutorial and caught an error when playing with it. Figured I'd submit a PR.

examples/blog/app.py
```diff
@@ -195,10 +195,13 @@ def index():
...
-                published=request.form.get('published') or False)
+            try:
+                entry = Entry.create(
+                    title=request.form['title'],
+                    content=request.form['content'],
+                    published=request.form.get('published') or False)
+            except IntegrityError:
+                flash('That title already exists.', 'warning')
             flash('Entry created successfully.', 'success')
             if entry.published:
```
coleifer: Assuming the integrity error was raised, what do you think is going to happen here?

coleifer:
I appreciate the intentions, but your change-set has more problems than it attempts to solve. I've fixed the underlying issue with 6a07c8e1d949724ae10c9e304b1d7d30214cf2c6

bdauer:
Oh shoot I didn't copy over my whole change. I had added an additional line after the flash to render_template('create.html') in my local code, forked and made a quick change in github's editor. But reading over your fix I suspect there's a good reason why that's also not the best option.

coleifer:
The reason it's not a good option (and whoops I need to fix something else, just remembered) is because you lose all your changes that you had been working on.

coleifer:
Oh actually nvm, I see that the template is pulling straight from request.form. Nevermind!

PR #1122, opened 2016-11-09
Fix pskel print statement for python3 compatibility

stratakis:
Invoking pskel under python3 produces an error due to the usage of print statement without parenthesis.


coleifer:
Whoops! Thank you.

PR #1126, opened 2016-11-10
Make pskel script installable under usr/bin/

stratakis:
Currently since pskel script does not have the .py extension, nothing happens when invoking setup.py install with that script. It just exists at the sources. The simplest way to make this script get installed is by adding it at the scripts sections in setup.py, thus making it installable under usr/bin/.


coleifer:
Thanks!

PR #1137, opened 2016-12-02
fix bad pg connections before returning to pool

brenguyen711:
For https://github.com/coleifer/peewee/issues/1138

Thanks!

### DIFF
--- a/playhouse/pool.py
+++ b/playhouse/pool.py
@@ -66,6 +66,11 @@ from peewee import MySQLDatabase
 from peewee import PostgresqlDatabase
 from peewee import SqliteDatabase
 
+try:
+    from psycopg2 import extensions as pg_extensions
+except ImportError:
+    pg_extensions = None
+
 logger = logging.getLogger('peewee.pool')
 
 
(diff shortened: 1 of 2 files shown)


playhouse/pool.py
```diff
@@ -142,6 +147,12 @@ def _connect(self, *args, **kwargs):
     def _is_stale(self, timestamp):
         return (time.time() - timestamp) > self.stale_timeout
 
+    def _is_ok(self, conn):
+        # Return True if it's safe to return this connection to the pool.
+        # Retrun False otherwise.
+        # Implementing classes are encouraged to fix the connection (if possible).
+        return True
```
coleifer: The API should probably not be extended in this way. Let's change the name of "_is_closed" to "_is_usable" and keep things to a single method.

coleifer:
This is fantastic, thank you. I will merge, but I'd like to tighten up the API first. Thanks again.

coleifer:
I've made some small cleanups/changes to the APIs but this feature is now merged. Thanks again.

082f13bf6b3a740c07bf35dafaf699e236b097e1

PR #1145, opened 2016-12-12
Remove join predicate requirement for cross joins.

cassidykelly:
I was able to run cross join queries after making this change.


coleifer:
Nice, thanks for the fix.

PR #1153, opened 2016-12-21
Add support for ON CONFLICT DO NOTHING in Postgres.

cassidykelly:
I saw in the documentation there is a plan to add support for Postgres 9.5's ON CONFLICT DO UPDATE (i.e. upsert), but I didn't see any mention of ON CONFLICT DO NOTHING. This is one simple way to support the latter, although it does not support adding a **_conflict_target_**.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2017,7 +2017,7 @@ class QueryCompiler(object):
         alias_map.add(model, model._meta.db_table)
         if query._upsert:
             statement = meta.database.upsert_sql
-        elif query._on_conflict:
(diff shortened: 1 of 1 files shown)


peerchemist:
Will this get merged? Any obvious show-stoppers?

coleifer:
Full support for ON CONFLICT is pretty complex. Take a look at this: https://www.postgresql.org/docs/9.5/static/sql-insert.html

coleifer:
This feels like a stop-gap implementation. I'd prefer t be comprehensive in implementing this feature, so I'm closing this.

PR #1173, opened 2017-01-13
Fixed setting of compound parentheses, fixes #1168

### LINKED ISSUE #1168: UNION with 3 or more queries leads to invalid SQL syntax (MySQL)
With the MysqlDatabase a UNION with 3 or more queries leads to a wrong syntax by wrapping individual UNION pairs in parentheses.
I don't know if other databases are affected as well, SQLiteDatabase seems to work.

Test Case:
```python
class A(Model):
    id = PrimaryKeyField()
    b = IntegerField()

    class Meta(object):
        database = MySQLDatabase(MagicMock())

class TestPeeweeUnion(TestCase):
    def test_union(self):
        union = (A.select() | A.select() | A.select())

        self.assertEqual('(SELECT `t1`.`id`, `t1`.`b` FROM `a` AS t1) UNION (SELECT `t2`.`id`, `t2`.`b` FROM `a` AS t2) UNION (SELECT `t3`.`id`, `t3`.`b` FROM `a` AS t3)', union.sql()[0])
```

The assert fails because instead the following string is generated:

```sql
((SELECT `t1`.`id`, `t1`.`b` FROM `a` AS t1) UNION (SELECT `t2`.`id`, `t2`.`b` FROM `a` AS t2)) UNION (SELECT `t3`.`id`, `t3`.`b` FROM `a` AS t3)
```

The first pair of UNION is in parentheses.

Affected Versions are 2.8.4 and 2.8.5.

Thank you for looking into this.

Edit: Judging by the test TestCompoundSelectSQL.test_multiple_with_parentheses this behavior is apparently desired. So in MySQL the parantheses should be set differently to the default.

danygielow:
This PR fixes the issue in #1168 by only wrapping non-compound queries in parentheses.

Changed the corresponding test as well.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1746,9 +1746,19 @@ class QueryCompiler(object):
         # We add outer parentheses in the event the compound query is used in
         # the `from_()` clause, in which case we'll need them.
         if node.database.compound_select_parentheses:
-            sql = '((%s) %s (%s))' % (l, node.operator, r)
-        else:
-            sql = '(%s %s %s)' % (l, node.operator, r)
+            # only add parentheses around non-compound queries, so that SQL syntax stays correct
+            if first_q._node_type != csq:
+                if inv:
+                    r = '(%s)' % r
+                else:
+                    l = '(%s)' % l
+            if second_q._node_type != csq:
+                if inv:
+                    l = '(%s)' % l
+                else:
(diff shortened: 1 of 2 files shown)


coleifer:
I fixed this using slightly more compact code, hope it's more readable now.

Issue #1189, opened 2017-02-10
SqliteQueueDatabase Writer.execute() error


krikunts:
Hello! I read your blogpost about SqliteQueueDatabase and want to try it out, but I have an error using SqliteQueueDatabase 
UnboundLocalError: local variable 'exc' referenced before assignment
https://github.com/coleifer/peewee/blob/d99f4a2649ce02da2f67769e381e51264f3c3eac/playhouse/sqliteq.py#L176
If there is exception during sql execution, the exc var is not visible outside of except block

coleifer:
What exception is underneath the `UnboundLocalError` I wonder? I would've thought that catching Exception would suffice but it looks like whatever error is being raised extends `BaseException` instead?

coleifer:
I couldn't replicate. I added a test-case in which invalid SQL is executed via the AsyncCursor and it all works as I'd expect. I've pushed a new release (2.8.6) in case the bug existed in the last release, however.

PR #1190, opened 2017-02-10
SqliteQueueDatabase exception fix for #1189

krikunts:
Fixing reference before assignment error #1189


coleifer:
No, `exc` is set in the `else` clause, which runs if no Exception is raised...

coleifer:
I was mistaken! In Python 3, the expression: `except <Exception> as <var>` does not keep `<var>` in scope beyond the exception block. So you were correct in raising this issue. It is fixed in `master` now.

krikunts:
Great! My bad, I didn't specify python version (it was 3.5, for the record)

PR #1195, opened 2017-02-19
Give TypeError if calling 'update', etc. from Model instance instead of from class

HaraldNordgren:
I hit a snag with this that took me hours to debug. Since these methods really only make sense as class methods, I don't you should to be able to call them from an instance.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4832,6 +4832,21 @@ class Model(with_metaclass(BaseModel)):
         for k, v in kwargs.items():
             setattr(self, k, v)
 
+    def __getattribute__(self, item):
+        classmethods = [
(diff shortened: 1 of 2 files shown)


coleifer:
This grosses up the code quite a bit. They already have the classmethod decorator.

HaraldNordgren:
I agree that it grosses up the code a bit. But it's very useful. Oh well, it's your repo.

coleifer:
The grossness isn't a reflection of your coding ability, I should make that clear -- it's just one of those problems that requires a kinda gnarly fix. Thanks for the patch, I'm sorry if it was wasted time on your part.

PR #1196, opened 2017-02-19
Give TypeError if calling 'update', etc. from Model instance instead of from class

HaraldNordgren:
Same as #1195 but based on the latest master commit with passing tests


HaraldNordgren:
It seems that Travis still builds from HEAD of origin/master with the PR commit placed on top, so this also fails.

HaraldNordgren:
Reverting "Move cdefs back into extern from block" and "Bring back tests" makes the tests work again. Again, this branch probably should not be merged.

But at least my commit is not breaking anything.

PR #1202, opened 2017-03-04
Make Model.get(val = None) work the same as Model.select().where(Model.val == None)

Necoro:
Currently `Model.select().where(val == None)` compiles into a `WHERE val IS NULL` clause. On the other hand `Model.get(val = None)` compiled into `WHERE val = NULL` and, if val is an FK field, tried to join.

With this PR, this is fixed and Model.get behaves just like Model.select().where()


coleifer:
Nice, thanks!

PR #1203, opened 2017-03-04
SQLite Closure Tables: Allow for many-to-many relations

Necoro:
Currently, the ClosureTable only allows for self-joins using the same table. But if you have many-to-many relations between one table using an intermediate table that was not expressible.

With this PR, a self-many-to-many-relation is now supported (as rare as it might be -- I needed it ;)).

I also took the liberty to update the documentation, but this probably could use a second look.

### DIFF
--- a/docs/peewee/playhouse.rst
+++ b/docs/peewee/playhouse.rst
@@ -773,6 +773,8 @@ sqlite_ext API notes
 
     :param model_class: The model class containing the nodes in the tree.
     :param foreign_key: The self-referential parent-node field on the model class. If not provided, peewee will introspect the model to find a suitable key.
+    :param referencing_class: The intermediate table for a many-to-many relationship.
+    :param id_column: For a many-to-many relationship: the originating side of the relation.
     :return: Returns a :py:class:`VirtualModel` for working with a closure table.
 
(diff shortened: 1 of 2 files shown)


coleifer:
Cool, I may make some small tweaks here and there, but looks great, thank you!

coleifer:
Actually, I've gone ahead and reverted this. It didn't seem to work when I started writing tests for it -- possibly I'm misunderstanding something?

Necoro:
Could you give me an example of what was not working as expected?

coleifer:
Ach, I take it back! I think I must've introduced a bug when I was editing the code to clean up some of the names.

coleifer:
Changes merged, I swapped the `id_column` parameter for one named `referencing_key` instead and made them `VirtualIntegerFields`. Other than that, changes are intact.

coleifer:
https://github.com/coleifer/peewee/blob/master/playhouse/sqlite_ext.py#L670-L758

Necoro:
Thanks. The fields `idcolumn` and `parentcolumn` are still strings, though. See the docs from closure.c:
```sql
SELECT element.* FROM element, ct1
WHERE element.groupid=ct1.id
AND ct1.root=?1
AND ct1.idcolumn='parentId'
AND ct1.parentcolumn='groupId';
```

coleifer:
Oh shit. Thanks! Will re-fix your fix.

coleifer:
04a3e022144ebf3db6a078ecd83f51dbbf766034

PR #1205, opened 2017-03-05
Make 'create_or_get' correctly support composite unique indexes and keys

### LINKED ISSUE #1204: create_or_get returns wrong instance
```python
>>> Tag.get(name='etf', prefix='wp')
<Tag (#16) wp:etf>
>>> Tag.create_or_get(name='etf', prefix='wp')
(<Tag (#1) mime:*>, False)
>>> Tag.get_or_create(name='etf', prefix='wp')
(<Tag (#16) wp:etf>, False)
```

See above, should be self explanatory. Going to analyze (and hopefully fix) it now.

Necoro:
This fixes #1204

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -33,6 +33,7 @@ from bisect import bisect_left
 from bisect import bisect_right
 from collections import deque
 from collections import namedtuple
+from collections import defaultdict
 try:
     from collections import OrderedDict
 except ImportError:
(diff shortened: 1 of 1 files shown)


coleifer:
Hmm... I received another pull-request with similar changes a while back and passed on it due to the complexity it introduced. My current thinking is that it'd be better to just remove `create_or_get` since it seems to cause confusion and the code one would need to write to work around it is very concise:

```python
try:
    return MyObj.create(whatever='foo')
except IntegrityError:
    return MyObj.get(MyObj.whatever == 'foo')
```

To me the result does not justify the means (and the subtlety involved). So I think in the next minor release I'll just remove the method instead.

PR #1206, opened 2017-03-06
Add interval field support for Postgres

alexlatchford:
I can add/amend the tests if you think this is a good approach.

PS. Sorry looks like my AutoPEP8 got in the way of the changes, I can remove if needs be.

### DIFF
--- a/docs/peewee/playhouse.rst
+++ b/docs/peewee/playhouse.rst
@@ -1309,6 +1309,7 @@ currently:
 * :ref:`full-text search <pg_fts>`
 * :py:class:`ArrayField` field type, for storing arrays.
 * :py:class:`HStoreField` field type, for storing key/value pairs.
+* :py:class:`IntervalField` field type, for storing durations.
 * :py:class:`JSONField` field type, for storing JSON data.
 * :py:class:`BinaryJSONField` field type for the ``jsonb`` JSON data type.
 * :py:class:`TSVectorField` field type, for storing full-text search data.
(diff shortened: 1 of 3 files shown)


coleifer:
What Python data-type is used by psycopg2 to represent an interval in Postgres? I'd assume a `datetime.timedelta`? I'm good with merging the changes (though please rebase into a single commit and remove the PEP8 changes).

alexlatchford:
Yep uses `timedelta` internally. Should be all handled transparently by `psycopg2` and the operations by the operators on `Node` from my initial tests.

Rebased as suggested and removed formatting.

coleifer:
Looks like the test model differs from the model used in the test-case itself, which is causing the builds to fail. You mind fixing that, and maybe adding an integration test that actually creates the table and verifies the data and data-types being stored/retrieved from the DB?

coleifer:
It was so close! Just needed to fix the tests up a bit. Any reason you closed this?

alexlatchford:
Unfortunately the project I needed this for was cancelled and I'm unlikely to get around to fixing this up anytime soon so didn't want to clutter things up. Feel free to take it on if you think it's a good addition :)

coleifer:
Thanks @alexlatchford -- I may do that when I've got a little free time. I appreciate the effort!

coleifer:
Committed f4ebd78b990efc6f0fc57b9a4d81f0cd6ecdf276

Issue #1219, opened 2017-03-12
CompositeKey in subquery not resolved


Necoro:
Given the following simple model with a composite PK:
```python
class DocumentTag(BaseModel):
    document = ForeignKeyField(Document, related_name = 'tags')
    tag = ForeignKeyField(Tag)

    class Meta:
        primary_key = CompositeKey('document', 'tag')
```

`DocumentTag.select()`  works as expected, yielding
```sql
SELECT "t1"."document_id", "t1"."tag_id" FROM "documenttag" AS t1
```

When I now want to use `DocumentTag` in a subquery, the CompositeKey is not resolved as one would expect (the query does not make sense, but works for illustration purposes):
`Document.select().where(DocumentTag.select() == 1)` now yields
```sql
SELECT "t1".*
FROM "document" AS t1 
WHERE ((SELECT ? FROM "documenttag" AS t2) = ?) 
[<peewee.CompositeKey object at 0x7f1c0465b438>, 1]
```

The same holds for other uses of subqueries, like joining with a subquery.

PR #1220, opened 2017-03-12
Use `get_primary_key_fields()` instead of `primary_key`, because the former knows how to handle composite keys

Necoro:
There is already a method `get_primary_key_fields()` which does just what is expected: return the list of all primary key fields, also handling CompositeKeys just right.

There are probably some more instances throughout the code, where using that method instead of `primary_key` would be advisable.

This should fix isse #1219.


coleifer:
Nice, thanks!

coleifer:
I also merged another fix that should also take care of any other instances of the issue.

PR #1242, opened 2017-04-07
Model generator support tables with no primary key

onlyice:
Currently for tables with no primary key, pwiz would generate model files without `primary_key = False`. Thus peewee assumes that there is a `id` column as primary key in the table, and the query failed.

Add `primary_key = False` in the `Meta` definition would prevent this error.


coleifer:
Merged equivalent patch.

PR #1250, opened 2017-04-15
Implement __ne__ for CompositeKey.

timster:
For a model with a single primary key, you can do both of these:

```python
(obj._meta.primary_key == obj._get_pk_value())
(obj._meta.primary_key != obj._get_pk_value())
```

And both return a `peewee.Expression` object.

However if the primary key happens to be CompositeKey, the `!=` compare returns `False`.

This fixes that. I can write tests if you'd like.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1578,6 +1578,11 @@ class CompositeKey(object):
                        for field, value in zip(self.field_names, other)]
         return reduce(operator.and_, expressions)
 
+    def __ne__(self, other):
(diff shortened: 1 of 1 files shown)


coleifer:
Wouldn't it just be:

```python
def __ne__(self, other):
    return ~(self == other)
```

timster:
Yeah, that works too. Forgot about `~` when I wrote it.

Issue #1262, opened 2017-05-05
Encryption with python3 throws TypeError


cym13:
Hi,

the AESEncryptedField in playhouse/fields.py is broken in python3.

Line 320 we see:

            key = key + bytes(' ') * (32 - len(key))

but bytes(' ') isn't valid python3 and will throw a TypeError.

coleifer:
This field has been removed in the latest release, 2.10.0. For details and explanation of why, see https://github.com/coleifer/peewee/releases/tag/2.10.0

PR #1263, opened 2017-05-05
Fix AESEncryptField for python3

ningirsu:
In python3 you can't create a bytes string with bytes(' ') because you need to specify an encoding.

```
2017-05-05 22:34:43,025 ERROR      File "/home/moutix/.virtualenvs/stepmania-server/lib/python3.5/site-packages/playhouse/fields.py", line 320, in get_cipher
    key = key + bytes(' ') * (32 - len(key))

2017-05-05 22:34:43,025 ERROR    Uncaught exception: string argument without an encoding
```

### DIFF
--- a/playhouse/fields.py
+++ b/playhouse/fields.py
@@ -317,7 +317,7 @@ if AES and Random:
         def get_cipher(self, key, iv):
             if len(key) > 32:
                 raise ValueError('Key length cannot exceed 32 bytes.')
-            key = key + bytes(' ') * (32 - len(key))
+            key = key + b' ' * (32 - len(key))
             return AES.new(key, AES.MODE_CFB, iv)
 
         def encrypt(self, value):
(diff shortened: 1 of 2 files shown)


ningirsu:
Related to: https://github.com/coleifer/peewee/issues/1262

ningirsu:
The tests for the playhouse extra_fields are not run automatically. Is there a particular reason?

playhouse/fields.py
```diff
@@ -317,7 +317,7 @@ def __init__(self, key, *args, **kwargs):
         def get_cipher(self, key, iv):
             if len(key) > 32:
                 raise ValueError('Key length cannot exceed 32 bytes.')
-            key = key + bytes(' ') * (32 - len(key))
+            key = key + b' ' * (32 - len(key))
```
coleifer: Why was this necessary?

coleifer:
Will all those byte literals be compatible with Python 2.6?

coleifer:
See comments on #1262 and #1264

PR #1272, opened 2017-05-12
Added note to quickstart regarding separate modules importing the db …

### LINKED ISSUE #1271: Tests in separate modules
I followed the quickstart guide and wasn't as thorough as I should have been with the docs, eventually locating the "extra" section which solved my problem. However, I think adding to the docs a best practices note in the quickstart on how to do tests with peewee would be beneficial.

My tests were in separate module, so I didn't know I was using a different db connection in my test modules when I made the db variable at the beginning of my test modules. I was trying to use transactions:

main/models.py
```
from peewee import *

db = SqliteDatabase('/home/ardeaf/my_app.db')

class BaseModel(Model):
    class Meta:
        database = db

class User(BaseModel):
    username = CharField(unique=True)
```

tests/test_models.py
```
from main.models import User

db = SqliteDatabase('/home/ardeaf/my_app.db')

test_save_user():
    db.atomic() as txn:
        username.Create("ardeaf")
        assert User.get(User.username == "ardeaf").username == "ardeaf"
        txn.rollback
````

So that wasn't working -- admittedly, I'm an idiot.  Tests passed, but database wasn't being rolled back. I eventually figured out the whole connection thing and ended up making my database dynamically defined at runtime.

However, it might be great to add a how-to on this in the quickstart docs in case someone was writing tests. I believe this would also be the case if you end up using the models in different modules, right? Perhaps a quick note explaining how to do that would help.

ardeaf:
Just a sample of what I was thinking of adding to the quickstart.

This commit fixes #1271.

### DIFF
--- a/docs/peewee/quickstart.rst
+++ b/docs/peewee/quickstart.rst
@@ -50,6 +50,9 @@ When starting a project with peewee, it's typically best to begin with your data
 .. note::
     Note that we named our model ``Person`` instead of ``People``. This is a convention you should follow -- even though the table will contain multiple people, we always name the class using the singular form.
 
+.. note::
+    When you define models this way, make sure that other modules importing it call the same "db" variable as the model's module: ``modulename.db``.  This way you will be able to use things like transactions or anything else that requires the same db connection as defined in Meta.
+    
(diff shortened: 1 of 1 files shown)


coleifer:
I don't think that this makes sense. What does it mean to "call" a db variable anyways? I think it's clear enough.

Issue #1287, opened 2017-05-31
model_to_dict - Support ManyToManyField


gcscaglia:
Consider the following models:

```python
class Tag(Model):
    name = TextField(unique=True)

class Gallery(BaseModel):
    title = TextField(null=True)
    tags = ManyToManyField(Tag, related_name='galleries')
```

When calling `model_to_dict` in a `Gallery` instance, one would expect the result dict to be:

```python
{
    'id': 1,
    'title': 'my title',
    'tags': [
        {
            'id': 1,
            'name': 'my tag'
        }
    ]
}
```

But the actual result is:

```python
{
    'id': 1,
    'title': 'my title',
    'gallerytagthrough_set': [
        {
            'id': 1,
            'tag': {
                'id': 1,
                'name': 'my tag',
            },
        },
    ],
}
```

While I do realize this is the result from how `ManyToManyField` works under the hood, the whole point of `ManyToManyField` is to pretend there is no relation table at all. Should't the returned dict be like the first one?

canhnht:
+1 for this

coleifer:
This is fixed in 3.0a by 0914cee. Example from test-case:

```python
class Owner(TestModel):
    name = TextField()

class Label(TestModel):
    label = TextField()

class Gallery(TestModel):
    name = TextField()
    labels = ManyToManyField(Label, backref='galleries')
    owner = ForeignKeyField(Owner, backref='galleries')

GalleryLabel = Gallery.labels.through_model

    def test_manytomany_field(self):
        data = (
            ('charlie', 'family', ('nuggie', 'bearbe')),
            ('charlie', 'pets', ('huey', 'zaizee', 'beanie')),
            ('peewee', 'misc', ('nuggie', 'huey')))
        for owner_name, gallery, labels in data:
            owner, _ = Owner.get_or_create(name=owner_name)
            gallery = Gallery.create(name=gallery, owner=owner)
            label_objects = [Label.get_or_create(label=l)[0] for l in labels]
            gallery.labels.add(label_objects)

        query = (Gallery
                 .select(Gallery, Owner)
                 .join(Owner)
                 .switch(Gallery)
                 .join(GalleryLabel)
                 .join(Label)
                 .where(Label.label == 'nuggie')
                 .order_by(Gallery.id))
        rows = [model_to_dict(gallery, backrefs=True, manytomany=True)
                for gallery in query]
        self.assertEqual(rows, [
            {
                'id': 1,
                'name': 'family',
[...]

PR #1288, opened 2017-06-01
#1287 - Support ManyToManyField in model_to_dict

gcscaglia:
Adds `dereference_many_to_many_fields` parameter to `model_to_dict` function.

This parameter controls whether to hide the junction table for `ManyToManyFields`.
See the related issue for details.

The default value is False, which preserves the old behavior for compatibility.

I've also added some simple tests for the new behavior; Let me know if they are insufficient and/or something should be changed.

### DIFF
--- a/playhouse/shortcuts.py
+++ b/playhouse/shortcuts.py
@@ -2,6 +2,7 @@ import sys
 
 from peewee import *
 from peewee import Node
+from playhouse.fields import ManyToManyFieldDescriptor
 
 if sys.version_info[0] == 3:
     from collections import Callable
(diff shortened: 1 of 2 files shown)


playhouse/shortcuts.py
```diff
@@ -134,6 +141,39 @@ def model_to_dict(model, recurse=True, backrefs=False, only=None,
...
+                        max_depth=max_depth - 1,
+                        dereference_many_to_many_fields=dereference_many_to_many_fields))
+
+                data[field_name] = accum
+
+            # The source foreign key MUST be excluded with or without backrefs
+            # to prevent the implicit backref from the junction table showing
+            # up the dict.
+            exclude.add(field_descriptor.src_fk)
+
```
coleifer: So much copy/paste :(

coleifer:
I've merged a patch that adds support manytomany in model_to_dict. Available in Peewee 3.x.

http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#model_to_dict

PR #1303, opened 2017-07-02
Modify add foreign key constraint.

EenTang:
Modify add foreign key constraint: support constraint type restrict and cascade


coleifer:
0137c5eb12172b1c127239f58ccbc98fbbbda982

PR #1310, opened 2017-07-06
Copied example code on 'Using' from database.rst to api.rst.

jbremson:
Simple doc change to clarify the use of 'Using' with example code.


coleifer:
Replaced with a similar patch. Thnx.

Issue #1336, opened 2017-08-22
Suport SSL connections


mgalvey:
At present, it appears that peewee / pwiz has no support for connecting to MySQL or Postgres using SSL. Such a feature would be useful in environments where the database is configured to only allow encrypted connections.

I'm not as familiar with Postgres, but I know that at least two modes should be available for MySQL: require SSL, and require SSL + verify CA.

coleifer:
Peewee operates at a higher-level than the database drivers, so any SSL support would need to be implemented at the `pymysql`/`psycopg2` layer instead of Peewee.

mgalvey:
It is implemented in PyMySQL (http://pymysql.readthedocs.io/en/latest/modules/connections.html?highlight=ssl), but I'm not sure how I would make use of that in pwiz.

I haven't tested, but it seems like the recommended way to do this in peewee proper would be to use connect_kwargs on instantiation of a Database object, correct?

coleifer:
You would just:

    db = MySQLDatabase('db_name', ssl='whatever')

Any keyword arguments that peewee doesn't recognize are passed back to the db driver, so `ssl` would get passed to pymysql in this case.

mgalvey:
Right. That makes sense for peewee proper and is in line with what I expected I'd have to do from your initial response.

Is there any way to do something similar with pwiz, or is that just something I'd have to make modifications to the code for? It looks from the current pwiz code like there's not an obvious way to get arbitrary options into the database constructor. Would a PR to add a convenience flag to pwiz (--ssl, for instance) be welcome?

coleifer:
Oh, I understand now what you're asking -- you may need to make some changes to pwiz or just write a wrapper script to do the same thing pwiz does.

azydevelopment:
@coleifer Also mentioning in this Issue. With the sensitive data we're dealing with and the MySQL DB that enforces SSL connections, we need a way to connect.

coleifer:
Same advise still applies:

> you may need to make some changes to pwiz or just write a wrapper script to do the same thing pwiz does.

What is with putting SSL on MySQL? Is your db exposed to the internet or something? Clustered across multiple data-centers?

azydevelopment:
So I'm not quite sure how pwiz works yet. Still working through the code and looking.

But to answer your question first, yeah, the DB is exposed to the internet. It's not clustered across multiple datacenters or anything but it does need to be accessed by multiple parties over the public internet.

[...]

PR #1342, opened 2017-08-26
Add --ssl option to pwiz

mgalvey:
Refs #1336 

Tested locally with both supported mysql drivers and didn't run into any problems. DID NOT test that psycopg tries SSL by default, but the docs for the client library that psycopg uses say that it does: https://www.postgresql.org/docs/9.1/static/libpq-ssl.html. From that page:

> The default value for sslmode is prefer.

This enables only the most simple ssl mode (with no CA verification), but it at least gets you around the problem of a server configured only to allow secure connections.

### DIFF
--- a/pwiz.py
+++ b/pwiz.py
@@ -39,6 +39,11 @@ def make_introspector(database_type, database_name, **kwargs):
         sys.exit(1)
 
     schema = kwargs.pop('schema', None)
+    ssl = kwargs.pop('ssl', False)
+    # According to postgres docs, ssl is attempted by default, so nothing
+    # should be needed for postgres.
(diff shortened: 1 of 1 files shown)


coleifer:
Going to pass on this, but thank you.

mgalvey:
Is there anything that would make this or a feature like it suitable for inclusion in pwiz, or is this use case just something you're not interested in supporting?

coleifer:
In 5...6 years you're the only person who's requested it :)

azydevelopment:
@coleifer I guess I should say that we really need this. I'm not sure how we can use peewee in a anything with even mildly sensitive data without SSL. Basically, it's for the exact reason that @mgalvey was saying.

PR #1347, opened 2017-09-08
Don't mark the column 'dirty' if there is no data change.

nad2000:
In order to reduce DB updates the columns should be marked if there was actually any data change...


coleifer:
The tests have exploded.

nad2000:
need more work on this one...

PR #1352, opened 2017-09-19
Allow releasing of stale connections back to pool.

sebastienvercammen:
Fine-grained control over the connection pool in non-web contexts can become more frustrating with the lightweight implementation because it's not implemented in peewee's core (e.g. after finishing a query, the connection isn't auto-released back to the pool), and in some structures context managers or decorators aren't desirable (requiring global access to the database's instance).

Adding this (very simple) option allows us to release connections back to the pool when they are stale rather than closing them entirely. This lets the user decide on how to manage what peewee considers stale connections.

There is one downside: the connection pool doesn't/can't know whether the query has finished so it's possible it will release busy connections back to the pool.

I hope this PR can serve as a discussion board for a better solution, because calling `Model._meta.database.close()` after every query or needing global access to the database's instance is not ideal.

### DIFF
--- a/playhouse/pool.py
+++ b/playhouse/pool.py
@@ -8,6 +8,9 @@ In a single-threaded application, only one connection will be created. It will
 be continually recycled until either it exceeds the stale timeout or is closed
 explicitly (using `.manual_close()`).
 
+Stale connections will either be closed (default) or will be released back to
+the pool (using `close_stale=False`).
+
 By default, all your application needs to do is ensure that connections are
 closed when you are finished with them, and they will be returned to the pool.
 For web applications, this typically means that at the beginning of a request,
@@ -23,6 +26,7 @@ Simple Postgres pool example code:
         'my_app',
         max_connections=32,
         stale_timeout=300,  # 5 minutes.
+        close_stale=True,  # Close stale connections.
         user='postgres')
 
     class BaseModel(Model):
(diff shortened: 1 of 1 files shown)


coleifer:
I'm not sure how a connection would still be busy, since peewee will block until the query is finished. Furthermore, connections aren't returned to the pool until you call `close()` on them.

If you don't wish for connections to be marked stale, you can also modify the `stale_timeout` parameter (which, by default, is not set).

sebastienvercammen:
@coleifer We've restructured the original application that brought us to this use case, but I submitted the PR in case there's a use for releasing connections to the pool rather than closing them.

You're correct that peewee blocks for queries, but I was thinking about what would happen if a different thread went through the connections and found one that passed `stale_timeout` while its original thread was still waiting for a response.

coleifer:
The connection wouldn't be available to another thread, as it would be "checked out" by the connection that was using it. There's a lock around connect/close, so no race condition either.

sebastienvercammen:
@coleifer I seem to be overlooking the locks around connect/close in `PooledDatabase`, could you link me to it?

coleifer:
They're in peewee's `Database` object.

sebastienvercammen:
@coleifer That makes an awful lot more sense, thanks for clarifying. If you see no use for the PR (which no longer fits our own use case), feel free to close it.

coleifer:
Glad to help

PR #1359, opened 2017-10-06
Database URL extension for retrieving endpoint information from AWS RDS

mguillau:
This pull-request adds an extension to playhouse's database URL mechanism to handle two new URL schemes, namely rds:// and rdsro:// (for rds "read-only"). These schemes simplify the use of AWS' Relational Database Service (RDS).

Through boto3 (the main python SDK for AWS), RDS can provide the host endpoint (rw or ro), its port, the master username and the database engine type (mysql, postgresql, aurora and mariadb supported for far in this patch) to any authorised AWS users, from just the database id (a case-insensitive string). With this information, it is then easy to build the actual database URL for peewee/playhouse to use for connecting.

For instance, one could provide the simple url "rds://test_db_user:test_db_pass@test_rds_db_id" which would automatically resolve to "mysql://test_db_user:test_db_pass@test_rds_db.ceqrivboeuyrbj.us-west-2.rds.amazonaws.com:5555/testdb" (assuming the RDS instance is mysql-compatible, listening on port 5555, etc.). The benefit is that the URL does not need to be changed as one manages its RDS database (changing the port, replacing hosts, adding replication and growing the fleet, etc.)

The main idea of the patch is to provide the function that converts a parsed rds:// or rdsro:// url into the corresponding parsed mysql:// or postgresql:// URL (dynamically determined depending on the engine type). The commit also provides a mechanism to add more such functions to resolve parsed URLs into new parsed URLs. Potential applications for this mechanism include reading database URLs or passwords from environment variables, configuration files, else other secret-sharing services (those applications are left for future pull requests).

Looking forward to comments and suggestions!

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -24,10 +24,24 @@ schemes = {
     'sqliteext+pool': PooledSqliteExtDatabase,
 }
 
+scheme_resolvers = {
+}
+
 def register_database(db_class, *names):
     global schemes
     for name in names:
         schemes[name] = db_class
+
+def register_resolver(url_resolver, *names):
+    global scheme_resolvers
+    for name in names:
+        scheme_resolvers[name] = url_resolver
+
+def resolve(parsed):
+    # could be made recursive with some safe-guards.
+    if parsed.scheme in scheme_resolvers:
+        parsed = scheme_resolvers[parsed.scheme](parsed)
+    return parsed
 
 def parseresult_to_dict(parsed):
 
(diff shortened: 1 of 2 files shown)


coleifer:
Because I'm not interested in maintaining this functionality, I'd suggest packaging this as a standalone project.

mguillau:
Thanks for the comments. I can easily re-submit a patch containing only the schema resolution part in db_url. Indeed, this would make it much easier to create standalone extensions as suggested. Would such a pull-request be of enough interest to you?

Issue #550, opened 2015-03-16
Foreign key enforcement in SQLite


wkschwartz:
Foreign keys are disabled by [default in SQLite3](http://www.sqlite.org/foreignkeys.html#fk_enable). It would be nice to have an optional `enable_foreign_keys` boolean argument to the `SqliteDatabase` constructor. It would have two effects. Suppose `db = SqliteDatabase(filename, enable_foreign_keys=True)`. Then

1. `db.foreign_keys is True`; and
2. `db.create_tables` would issue

```SQL
PRAGMA foreign_keys = ON;
```

Issuing that `PRAGMA` in the middle of a multi-statement transaction has no effect and swallows the error. It is only possible to turn on `foreign_keys` in SQLite versions greater than 3.6.19 when SQLite is compiled with `SQLITE_OMIT_TRIGGER` and `SQLITE_OMIT_FOREIGN_KEY ` turned _off_. The versions of SQLite compiled into Python versions 2.7.9 and 3.4.3 on Windows appear to support `foreign_keys`.

coleifer:
You can subclass `SqliteDatabase` and override the `_add_conn_hooks` method. I don't think I will add this as a database option, but it should be easy to add yourself.

wkschwartz:
Using `_add_conn_hooks` doesn't work. In fact, it doesn't even work for `journal_mode`. See #554.

I created a wrapper connection function for my global database object `_db` that looks something like:

```python
def connect():
    _db.connect()
    if isinstance(_db, peewee.SqliteDatabase):
       if _db.execute_sql('PRAGMA foreign_keys).fetchone() is None:
          _db.close()
          raise RuntimeError('Your installed version of SQLite does not support foreign keys.')
       _db.execute_sql('PRAGMA foreign_keys = ON')
```

coleifer:
I've gone ahead and addressed the issue in #554. I'm not sure why `_add_conn_hooks` won't work. Can you not write:

```python
cursor = conn.cursor()
cursor.execute('pragma foreign_keys=ON')
```

wkschwartz:
I'm sure that would work. I just gave up too quickly.

Does the fact that you recommend `_add_conn_hooks` mean you intend on keeping its API stable? I assumed from the fact that it's a private method client code (such as mine) shouldn't use it.

coleifer:
I think it'd be good to add a public API for interacting with a new connection. I will add that, add docs, and update you with my progress.

coleifer:
[...]

PR #1366, opened 2017-10-17
Add clear explanations about the implications of SQLite having foreign keys off by default, in combination with default PrimaryKeyField ID reuse.

bryhoyt:
To address the problems raised in #550


coleifer:
Very nice, thank you!

PR #1368, opened 2017-10-20
Add support for using verbose names as CSV header values

brandond:


coleifer:
CSV module is no longer included in Peewee as of 3.0.0.

brandond:
@coleifer is there a suggested replacement? This was a great simple way to dump stuff out.

coleifer:
The dumping code is just a thin wrapper: https://github.com/coleifer/peewee/blob/stable/playhouse/csv_utils.py

You can easily backport it into your codebase.

PR #1389, opened 2017-11-17
Do not force primary_key value on returning clause

gdraynz:
Using Postgresql, when calling `.returning(*)` on an `InsertQuery`, the `RETURNING` clause was always set to my single primary_key.
I simply inverted the two possible returning clauses to ensure `.returning()` takes the priority.


coleifer:
Cheers, thanks!

PR #1412, opened 2017-12-29
Document exception hierarchy

zmwangx:
Sometimes I want to make sure I'm catching all possible peewee exceptions, but not knowing the base exception makes it hard. Having the exception hierarchy in API docs (inspired by [docs for built-in exceptions][1]) could be handy for this.

[1]: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

### DIFF
--- a/docs/peewee/api.rst
+++ b/docs/peewee/api.rst
@@ -2717,6 +2717,22 @@ Metadata Types
 
         The name of the table containing this foreign key.
 
+Exception Hierarchy
+-------------------
+
+::
+
+  PeeweeException
+   +-- DatabaseError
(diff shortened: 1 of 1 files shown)


coleifer:
Cheers!

zmwangx:
Wow, that was fast. Thanks.

coleifer:
For what it's worth, the peewee exceptions are based on those already documented in the DB-API 2.0 spec. https://www.python.org/dev/peps/pep-0249/#exceptions

zmwangx:
Good to know, thanks for the pointer.

PR #1413, opened 2017-12-30
Update metadata.rst

sigma65535:
from the attributes table of Meta ,
the property of Meta “db_table” should be “table_name”。


coleifer:
Documentation changes that pertain only to 3.0a are contained in the 3.0a branch.

PR #1436, opened 2018-01-29
Add base class rename

aconz2:


coleifer:
Cheers. This wasn't exactly a public API, but I guess it doesn't hurt to include it?

aconz2:
Yea I just ran into this while upgrading some code so figured I would try to give others a heads up. I have some scripts that do consistency checks that totally reach in where they're not supposed to.

Thank you for your work!

PR #1463, opened 2018-02-02
Pass kwargs through Function decorator. Fixes #1462

### LINKED ISSUE #1462: fn.ARRAY_AGG in select query tries to re-cast result
The `ARRAY_AGG` function combines the result of multiple row in a single ArrayField. Peewee doesn't expect that and tries to re-cast to the original field type. For example:

``` python
Network.select(fn.ARRAY_AGG(Network.id)).get()
```
Which returns a int[] in PostgreSQL, but peewee tries to recast to int
``` python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/working/QLKNN-develop/NNDB/model.py in <module>()
----> 1 Network.select(fn.ARRAY_AGG(Network.id).alias('wut')).get()

~/working/peewee/peewee.py in get(self)
   5429         clone._cursor_wrapper = None
   5430         try:
-> 5431             return clone.execute()[0]
   5432         except IndexError:
   5433             sql, params = clone.sql()

~/working/peewee/peewee.py in __getitem__(self, item)
   3338             return self.row_cache[item]
   3339         elif isinstance(item, int):
-> 3340             self.fill_cache(item if item > 0 else 0)
   3341             return self.row_cache[item]
   3342         else:

~/working/peewee/peewee.py in fill_cache(self, n)
   3383         while not self.populated and (n > self.count):
   3384             try:
-> 3385                 iterator.next()
   3386             except StopIteration:
   3387                 break

~/working/peewee/peewee.py in next(self)
   3439             obj = self.cursor_wrapper.row_cache[self.index]
   3440         elif not self.cursor_wrapper.populated:
-> 3441             self.cursor_wrapper.iterate()
   3442             obj = self.cursor_wrapper.row_cache[self.index]
   3443         else:

~/working/peewee/peewee.py in iterate(self, cache)
   3361             self.initialized = True
   3362         self.count += 1
-> 3363         result = self.process_row(row)
   3364         if cache:
   3365             self.row_cache.append(result)

~/working/peewee/peewee.py in process_row(self, row)
   5913 
   5914     def process_row(self, row):
-> 5915         data = super(ModelObjectCursorWrapper, self).process_row(row)
   5916         if self.is_model:
   5917             # Clear out any dirty fields before returning to the user.

~/working/peewee/peewee.py in process_row(self, row)
   5876             attr = fields[i].name if fields[i] is not None else columns[i]
   5877             if converters[i] is not None:
-> 5878                 result[attr] = converters[i](row[i])
   5879             else:
   5880                 result[attr] = row[i]

~/working/peewee/peewee.py in python_value(self, value)
   3588 
   3589     def python_value(self, value):
-> 3590         return value if value is None else self.coerce(value)
   3591 
   3592     def get_sort_key(self, ctx):

TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'
```
Can this be worked around (for example by using the equivalent of `AsIs`) or fixed?

Karel-van-de-Plassche:
This enables the user to do for example

``` python
fn.ARRAY_AGG(Some.thing, coerce=False)
```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1225,8 +1225,8 @@ class Function(ColumnBase):
             self._coerce = coerce
 
     def __getattr__(self, attr):
-        def decorator(*args):
-            return Function(attr, args)
+        def decorator(*args, **kwargs):
+            return Function(attr, args, **kwargs)
         return decorator
 
     def over(self, partition_by=None, order_by=None, start=None, end=None,


coleifer:
I've restricted the kwargs to just "coerce" and merged a similar patch. Thank you for pointing this out: 51cb6bdc25f62916fbc1f9bc471f1a1bccc1a7da

PR #1465, opened 2018-02-02
Fix not_in with empty list

aconz2:
Using `not_in` with an empty list produces malformed SQL

```python
import peewee
from peewee import *

class MyBase(Model):

    class Meta:
        database = SqliteDatabase(':memory:')

class Foo(MyBase):
    name = CharField()

print(peewee.__version__)
print(Foo.select().where(Foo.name.not_in([])).sql())
```

```
2.10.2
('SELECT "t1"."id", "t1"."name" FROM "foo" AS t1 WHERE ("t1"."name" NOT IN ())', [])
```
```
3.0.10
('SELECT "t1"."id", "t1"."name" FROM "foo" AS "t1" WHERE ("t1"."name" NOT IN )', [])
```

I think (?) that is the right fix. And added a new test to cover it

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1357,7 +1357,10 @@ class NodeList(ColumnBase):
     def __sql__(self, ctx):
         n_nodes = len(self.nodes)
         if n_nodes == 0:
-            return ctx
+            if self.parens:
+                return ctx.literal('()')
+            else:
+                return ctx
         with ctx(parentheses=self.parens):
             for i in range(n_nodes - 1):
                 ctx.sql(self.nodes[i])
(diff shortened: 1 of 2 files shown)


coleifer:
Postgresql treats `IN ()` (and incidentally `NOT IN ()`) as invalid SQL expressions. Because the semantics of `IN ()` are equivalent to false, Peewee replaces `IN ()` with `1 = 0`. I suppose it should do similar rewriting for `NOT IN ()`, converting it to `1 = 1` since it is always true.

That being said, it seems like the proposed change may be correct.

coleifer:
Replaced your patch with bc61d3fa515c462cb86714e00ffdf728121a90f7 -- which handles the IN/NOT IN special-case correctly. I **did**, however, choose to keep your changes to `NodeList.__sql__` as I think they're semantically correct.

PR #1496, opened 2018-02-15
Create schema when 'creating all' if it doesn't exist

Karel-van-de-Plassche:
As it says in the table. Not sure if we want to use the safe version of this method in `_create_table` itself, which is also possible. Also, I was not sure where to put Unit Tests for this. If you give me a location, I can write them if you want.

PS. One test fails for PostgreSQL, but it also fails without my pull request.

``` python
FAIL: test_db_table (tests.schema.TestModelDDL)
  File "/home/karel/working/peewee/tests/schema.py", line 126, in test_db_table
    'CREATE TABLE "A_tbl" ("id" INTEGER NOT NULL PRIMARY KEY)'])
  File "/home/karel/working/peewee/tests/schema.py", line 83, in assertCreateTable
    self.assertEqual([sql] + indexes, expected)
AssertionError: Lists differ: ['CREATE TABLE "A_tbl" ("id" SERIAL NOT NULL PRIMARY KEY)'] != ['CREATE TABLE "A_tbl" ("id" INTEGER NOT NULL PRIMARY KEY)']

First differing element 0:
'CREATE TABLE "A_tbl" ("id" SERIAL NOT NULL PRIMARY KEY)'
'CREATE TABLE "A_tbl" ("id" INTEGER NOT NULL PRIMARY KEY)'

- ['CREATE TABLE "A_tbl" ("id" SERIAL NOT NULL PRIMARY KEY)']
?                              ^  ---

+ ['CREATE TABLE "A_tbl" ("id" INTEGER NOT NULL PRIMARY KEY)']
?                              ^^^^^


```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4426,6 +4426,18 @@ class SchemaManager(object):
     def _create_context(self):
         return self.database.get_sql_context(**self.context_options)
 
+    def _create_schema(self, safe=True):
+        meta = self.model._meta
+        ctx = self._create_context()
+        ctx.literal('CREATE SCHEMA ')
+        if safe:
+            ctx.literal('IF NOT EXISTS ')
+        ctx.literal(self.model._meta.schema)
+        return ctx
+
+    def create_schema(self, safe=True):
+        self.database.execute(self._create_schema(safe=safe))
+
     def _create_table(self, safe=True, **options):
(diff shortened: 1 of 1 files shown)


coleifer:
I am going to pass on this, but thanks for the patch.

PR #1498, opened 2018-02-15
Fixed testcases broken by bugfix 9edf260

Karel-van-de-Plassche:
I think this test has to be changed, as primary keys are an AutoField, and Postgress' AutoField is SERIAL. This expected behaviour somehow only surfaces after 9edf260.


Karel-van-de-Plassche:
For sqlite the autofield IS an INTEGER, so this solves nothing. Opened issue #1499

PR #1506, opened 2018-02-21
Model.table_exists() now checks whether table exists in its schema

skyhound:
Model.table_exists() with safe=True was checking whether table exists in default schema('public') instead of Model schema.


coleifer:
Thank you!

skyhound:
You're welcome. Great ORM, btw.

PR #1528, opened 2018-03-08
Clarify comment that sounded like create_tables() in Peewee 3 will dr…

bryhoyt:
…op your table if it exists.

This scared me on first read. I had to read it a couple times to figure out that the second part of the parenthetical comment was talking about the drop_tables() method. And I had to read the source code to be fully sure.

I like the updated functionality, though. It's a great improvement, thanks!


coleifer:
Cheers!

PR #1531, opened 2018-03-09
Support prefetching ManyToManyFields

iBelieve:
Previously, prefetching a relationship exposed in a ManyToMany field would only expose the through model’s prefetched data. Now, the many to many field correctly exposes the prefetched data via the through field’s prefetched results, rather than requerying what has already been prefetched.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4246,16 +4246,19 @@ class ManyToManyFieldAccessor(FieldAccessor):
         self.src_fk = self.through_model._meta.model_refs[self.model][0]
         self.dest_fk = self.through_model._meta.model_refs[self.rel_model][0]
 
-    def __get__(self, instance, instance_type=None):
+    def __get__(self, instance, instance_type=None, force_query=False):
         if instance is not None:
-            return (ManyToManyQuery(instance, self, self.rel_model)
-                    .join(self.through_model)
-                    .join(self.model)
-                    .where(self.src_fk == instance))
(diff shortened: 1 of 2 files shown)


coleifer:
Interesting... Since the `prefetch` API requires the specification of the intervening models, it *does* seem a bit un-intuitive to access the related instances through the many-to-many field. Or maybe it is intuitive? I'm not sure...ManyToManyField is such a hack!

iBelieve:
At least for my use case, it feels intuitive, as I'm trying to fetch a list of users and their teams. I'm using prefetch as an optimization to prevent an N+1 situation for fetching teams for each user, so it feels most intuitive to access prefetched teams through the ManyToManyField.

Basically, I'd like to go from this:

```
for user in User.select():
    print(user.teams)
```

to the optimized version:

```
for user in prefetch(User.select(), TeamMember, Team):
    print(user.teams)
```

Without this, I'd have to do something like:

```
for user in prefetch(User.select(), TeamMember, Team):
    teams = [through.team for through in user.teammember_set]
    print(teams)
```

Which feels un-intuitive and requires that I treat prefetched queries differently from regular queries. Of course, the alternative is do not use ManyToManyField and just have a property on user that fetches teams which also works with prefetched, but for my use case ManyToManyField is simple and meets my needs.

coleifer:
Agreed, thanks for the patch!

PR #1535, opened 2018-03-11
Add big integer auto field support.

delobanov:
I added a new class that represents 8 bit autoincrement field. Tested this on postgresql.


coleifer:
Replaced with cleaner implementation in 37e538f. Thanks.

PR #1541, opened 2018-03-16
Make connecting signals operation idempotent

bmwant:
## Rationale
Attaching receivers should not crash a program. 

## Use cases
1. Module has been reloaded
`m.py`
```
from peewee import *
from playhouse import signals

class A(signals.Model):
    a = TextField(default='')

state = []

@signals.post_save(sender=A)
def post_save(sender, instance, created):
    state.append((sender, instance, instance._pk, created))
```
and `test.py`
```
import m

try:
    from importlib import reload
except ImportError:
    pass

reload(m)
```
will raise `ValueError`.

2. We should allow assigning same receiver multiple times (it should be idempotent) to handle cases like this
`test2.py`
```
from peewee import *
from playhouse import signals


class A(signals.Model):
    a = TextField(default='')


database = SqliteDatabase(':memory:')
A._meta.set_database(database)
A.create_table()

state = []


@signals.post_save(sender=A)
def post_save(sender, instance, created):
    state.append((sender, instance, instance._pk, created))


def ensure_receivers_are_connected():
    # I am not sure whether `flush` was invoked or not at some point
    signals.post_save(sender=A)(post_save)


ensure_receivers_are_connected()

a = A(a='my text')
a.save()
assert len(state) == 1
```
But I guess we still need to prohibit binding in a such way function to different sender. In case you intentionally want to do such a thing you have `disconnect` method.

### DIFF
--- a/playhouse/signals.py
+++ b/playhouse/signals.py
@@ -2,6 +2,7 @@
 Provide django-style hooks for model events.
 """
 from peewee import Model as _Model
+from peewee import logger
 
 
 class Signal(object):
@@ -13,8 +14,13 @@ class Signal(object):
         if name not in self._receivers:
             self._receivers[name] = (receiver, sender)
             self._receiver_list.append(name)
+        elif (receiver, sender) == self._receivers[name]:
+            logger.warning('receiver %s has been already connected' % name)
         else:
-            raise ValueError('receiver named %s already connected' % name)
+            raise ValueError(
+                'receiver %s is already attached to another sender %s' %
+                (name, self._receivers[name][1])
+            )
 
     def disconnect(self, receiver=None, name=None):
         if receiver:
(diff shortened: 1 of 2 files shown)


bmwant:
@coleifer could you review this one?

coleifer:
I don't think this is exactly a bug, and I don't love the solution proposed in this patch. I don't really think of registering a signal as an idempotent operation, and registering the same combination of receiver/name/sender seems like a user mistake.

I've tried to address some of the quirks of connect/disconnect in two commits:

* 8992c69ed8708d89e4f888d97b997a0232098141
* 2d421a9db9ddc6223c35972291afe37bc2b20c68

Registering a handler twice for the same combination of receiver/name/sender will result in a `ValueError`. I'm not clear on how this would come up in practice, since when a module is imported once it continues to remain in the sys module cache. If you want to avoid side-effects you can do your registration in an init function that only runs once.

PR #1562, opened 2018-04-06
Update api.rst for Sqlitedatabase inheritance

Declan1984:
Adding an explicit indication that Sqlitedatabase inherits from Database; if proposed change is accepted, I can go into the other classes and add similar explicit reference.


coleifer:
Thanks I'll pass on this for now. It'd be nice if the python domain of ReST supported this type of annotation but I couldn't find it.

PR #1588, opened 2018-04-26
'Metadata' object has no attribute 'model_ref'

klen:
Fixed mistype in Metadata class. It breaks migrations for now.


coleifer:
Thank you!

PR #1623, opened 2018-06-06
Add an option to try to guess foreign keys

etiennepouliot:
Sometimes the foreign keys constraint arent't defined in the databases. (old version of RoR for example).

With this options, it will try to guess foreign keys with the name of the tables and an "_id" suffix.


coleifer:
I'm not interested in this patch, but thanks.

Issue #1660, opened 2018-07-05
the order of importing mysql database driver is wrong


earlzo:
```python
try:
    import pymysql as mysql
except ImportError:
    try:
        import MySQLdb as mysql  # prefer the C module.
        mysql_passwd = True
    except ImportError:
        mysql = None
```
As your comment, we should prefer the C module as default because it has better performance for production but the order of importing mysql database driver is wrong.

It should be great If we can choose our database driver like sqlalchemy.

coleifer:
So `mysql-python` (`MySQLdb`) does not support Python 3, which is why it is no longer preferred. The comment needs to be updated.

If you would like to explicitly use a particular driver, it is quite easy:

```python
import MySQLdb

class MySQLDatabaseCustom(MySQLDatabase):
    def _connect(self):
        return MySQLdb.connect(db=self.database, **self.connect_params)
```

earlzo:
[farcepest/MySQLdb1](https://github.com/farcepest/MySQLdb1)  has already been abandoned, [PyMySQL/mysqlclient-python](https://github.com/PyMySQL/mysqlclient-python) has perfect Python3 support without losing performance benefit.

coleifer:
The code snippet I shared should work just fine. Again, peewee will continue to prefer pymysql, but you can always subclass and override the `_connect()` method. It's a bit surprising that you have both drivers installed, as you don't wish to use pymysql.

coleifer:
No disturbance -- I recognize it's a somewhat opinionated stance and the comment was incorrect, which made it worse. Hope you're able to get things working.

PR #1661, opened 2018-07-05
Fixed: import mysql driver

earlzo:
Fixed: #1660


coleifer:
As I mentioned, I intend to prefer pymysql as it supports Python 3.x (whereas MySQLdb does not). I've updated the comment and provided a code snippet you can use to explicitly use MySQLdb (when you have both drivers installed for whatever reason).

PR #1665, opened 2018-07-11
Improve imports in shortcuts extension

gustavi:
Improve the imports in this file.

First statement use wildcard import (and this is bad) and second import something already imported.

Import only required objects is the good way to do.

### DIFF
--- a/playhouse/shortcuts.py
+++ b/playhouse/shortcuts.py
@@ -1,7 +1,6 @@
 import sys
 
-from peewee import *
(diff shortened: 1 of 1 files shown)


coleifer:
Peewee's main module `peewee.py` defines a module-level `__all__` which controls what is exported when you `import *`. Peewee uses the convention of `import *` internally, so I am going to pass on this change, even though I do understand your reasoning behind the suggestion.

PR #1670, opened 2018-07-14
Added support for introspecting views for Postgresql.

NCarson:
A command line switch will introspect views instead of tables in pwiz.py. A
get_views function was added for Postgres and NotImplemented version for the
interface. It would probably be more convenient for users to do both tables and
views at once, but this was an extremely simple way to do it. I did not add it for
others since I am mostly familiar with Postgres, but it seems pretty easy.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2768,6 +2768,9 @@ class Database(_callable_context_manager):
     def get_tables(self, schema=None):
         raise NotImplementedError
 
+    def get_views(self, sechema=None):
+        raise NotImplementedError
+
     def get_indexes(self, table, schema=None):
         raise NotImplementedError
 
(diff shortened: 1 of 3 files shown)


coleifer:
I'm going to pass on this for now, but may reconsider adding this functionality (if it can be added cleanly for all supported database implementations). Thank you -- I may end up implementing this at some point and your code will certainly come in handy.

coleifer:
Also, for what it's worth, if you want to increase the likelihood of getting a "new feature" patch merged, it should have tests and documentation as well.

coleifer:
This is actually now implemented by 273008413737ff16d26b27f846a315cdf4875615

Issue #262, opened 2013-11-27
Type/Range Checking Input?


twodayslate:
Is it possible to do range checking on an insert? For example, when inserting a float it must be between 1.0 and 5.0

PR #1682, opened 2018-07-22
Update to Maintain Compatibility

jseagrave21:
This change allows compatibility between projects formerly using peewee2.6.2.

When trying to use the latest version of peewee we get the following errors relating to `_StringField`: "Could not get key 'utf-8' codec can't decode byte 0xb4 in position 3" or similar.

Making this change allows easy compatibility.

See issue #262 in neo-python project here: https://github.com/CityOfZion/neo-python/issues/262

### DIFF
--- a/Update to Maintain Compatibility
+++ b/Update to Maintain Compatibility
@@ -4023,7 +4023,10 @@ class _StringField(Field):
         if isinstance(value, text_type):
             return value
         elif isinstance(value, bytes_type):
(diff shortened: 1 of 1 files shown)


coleifer:
If you're storing a byte array -- i.e., *not* UTF8-encoded text -- you should be using a `BlobField`.

jseagrave21:
@coleifer Thank you for your help!

PR #1702, opened 2018-08-21
Reflection: Fix broken SQL statement for text-typed default values.

stlehmann:
When using the refection module on text-typed (unicode on py2, str on py3) Columns with default values there were parantheses missing around the generated SQL statement. That leads to errors on model migration.

I fixed this error by checking the data type of the default value and adding parantheses around it if it is a text type.

### DIFF
--- a/playhouse/reflection.py
+++ b/playhouse/reflection.py
@@ -19,6 +19,12 @@ try:
 except ImportError:
     postgres_ext = None
 
+import sys
+if sys.version_info[0] == 2:
+    text_type = unicode
+else:
+    text_type = str
+
 RESERVED_WORDS = set([
     'and', 'as', 'assert', 'break', 'class', 'continue', 'def', 'del', 'elif',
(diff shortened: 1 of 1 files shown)


coleifer:
You mean quotation marks, not parentheses...

coleifer:
I think the better way to fix this would be to introspect the field type rather than the default value.

stlehmann:
You are right ... "quotation marks" it is. 

I decided to check the type of the default value because even if you have a certain numeric field type the default value could be any type depending on the users input. E.g. the following is certainly possible:

    import peewee as pw
    class A(pw.Model):
        name = pw.CharField(default=1)
        num = pw.IntegerField(default="foo")

I know it makes no sense but peewee allows to set defaults that don' t match the field type. Maybe that should get checked, too. But to prevent the SQL-statement from breaking it seams better to me to check the type of the default value.

PR #1720, opened 2018-09-11
Connect to DB only if not connected already

mekanix:
When testing Flask apps, failure is "peewee.OperationalError: Connection already opened." I used the same logic inside connect_db as the one used in close_db.

### DIFF
--- a/playhouse/flask_utils.py
+++ b/playhouse/flask_utils.py
@@ -178,7 +178,8 @@ class FlaskDB(object):
         return self._model_class
 
     def connect_db(self):
-        self.database.connect()
+        if self.database.is_closed():
+            self.database.connect()
 
     def close_db(self, exc):
         if not self.database.is_closed():


coleifer:
I'll pass, because in my opinion an unclosed connection is a bug and this could lead to sloppy programming. You are welcome to subclass this in your own app and override the method as you like, however.

mekanix:
Maybe I wasn't clear enough, so I apologize. If you use FlaskDB with `sqlite:///:memory:`, there is no test you can write that won't break with mentioned error, so if there is a bug, it's on PeeWee side. Other DBs are fine and don't throw an error. Subclassing is OK, but I think it would be better if it would work without the subclass. I can write the simplest showcase, if needed. Also, if you think finding why multiple connections are opened, that's fine too, but I wonder why the same check is OK in `close_db` but isn't in `connect_db`?

coleifer:
> If you use FlaskDB with sqlite:///:memory:

Well, yes -- of course this will not work because every time you connect or close the connection, the data goes away. If you want to use an in-memory sqlite database, then obviously FlaskDB won't work for you.

The check is not present in connect_db() because an already-opened connection at that point indicates a dangling / unclosed connection from somewhere in the app.

mekanix:
OK, I didn't know one should not use FlaskDB with :memory: for tests. What would be the appropriate way to have sqlite in dev, sqlite:///:memory: in test and postgresql in production? As I already have different config class for each of them, should I use `SqliteDatabase` and `PostgresqlDatabase` classes in config? I did try closing DB in tearDown, but I still get this error. So, to put it short, how should I use sqlite:///:memory: for tests with pytest or nose?

PR #1719, opened 2018-09-10
Feature/manytomanyfield-ondelete

Susensio:
#1718 
IDK if I'm creating the pull request correctly. I have run tests and changes do not throw any errors.

PR #1724, opened 2018-09-13
Only validate if not backref

Susensio:
This validation was throwing an exception when creating the backref, because it has a through_model and also receives on_delete and on_update values.

#1719


coleifer:
Thank you

Issue #1730, opened 2018-09-20
VARCHAR not correctly parsed in pwiz


jbriales:
Hi,
I was using pwiz in python3 with sqlite and it seems VARCHAR fields are not correctly parsed:

If no max length is provided, the field is not recognized:
```
CREATE TABLE COMPANY(
   ID INT PRIMARY KEY     NOT NULL,
   NAME           VARCHAR NOT NULL,
   AGE            INT     NOT NULL,
   ADDRESS        TEXT,
   SALARY         REAL
);
```
```
class Company(BaseModel):
    address = TextField(db_column='ADDRESS', null=True)
    age = IntegerField(db_column='AGE')
    id = PrimaryKeyField(db_column='ID')
    name = UnknownField(db_column='NAME')  # VARCHAR
    salary = FloatField(db_column='SALARY', null=True)

    class Meta:
        db_table = 'COMPANY'
```

If max length is provided, I don't get it in models.py:
```
CREATE TABLE COMPANY(
   ID INT PRIMARY KEY     NOT NULL,
   NAME           VARCHAR(50) NOT NULL,
   AGE            INT     NOT NULL,
   ADDRESS        TEXT,
   SALARY         REAL
);
```
```
class Company(BaseModel):
    address = TextField(db_column='ADDRESS', null=True)
    age = IntegerField(db_column='AGE')
    id = PrimaryKeyField(db_column='ID')
    name = CharField(db_column='NAME')
    salary = FloatField(db_column='SALARY', null=True)

    class Meta:
        db_table = 'COMPANY'
```

coleifer:
Thanks, I've fixed the issue where VARCHAR without a length constraint is now recognized correctly with Sqlite -- fdeb92c1639c1dc9301ae501b95e36d83a63ee74

Regarding the max_length parameter, this has no effect in Sqlite and so it is fine to ignore it:

```sql
sqlite> create table kv (key varchar(4) not null primary key, value varchar(8) not null);
sqlite> insert into kv (key, value) values ('abcdefghij', '12345678901234567890');
sqlite> select * from kv;
abcdefghij|12345678901234567890
```

jbriales:
Thanks for the fix, it's working now :)
However I still found two failing cases in my database:
# CHAR[38]
# FLOAT
I will make a pull request next.

coleifer:
Eh.. You know SQLite only supports the following data-types, right?

* TEXT
* BLOB
* INTEGER
* REAL
* NULL

[...]

PR #1731, opened 2018-09-21
Match FLOAT  and CHAR[...] in sqlite reflection.

jbriales:
Fix issue related to https://github.com/coleifer/peewee/issues/1730#issuecomment-423461785


coleifer:
So SQLite only supports the following data-types:

* TEXT
* BLOB
* INTEGER
* REAL
* NULL

I think I'm going to pass on the square bracket change, but will add support for float along with some others:

d31e7bf972e69e227c10694fac71b966959e715f

PR #1753, opened 2018-10-15
Update models.rst

Walicen:
Allow null values in the fields


coleifer:
This is incorrect. Fields are non-null by default. So the documentation shows the field parameter with its default value, which is False.

Walicen:
OH Sorry!

PR #1776, opened 2018-11-04
PostgreSQL UPDATE statement with FROM clause with subquery support

nilp0inter:
This is a second try on #1775 , because the request was accepted and then reverted I am opening this.

The changes with master are:

1. I reverted the SET part in the UPDATE test queries.
2. I wrote a helper function to recursively qualify only ColumnBase member of Expressions.
3. I applied this helper to the RHS of the assignments in the SET part of UPDATE queries.

All tests, including the new regressions tests appear to be passing now.

Let me know what you think.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -387,6 +387,19 @@ class _callable_context_manager(object):
         return inner
 
 
+def qualify_expression(node):
+    if isinstance(node, Expression):
+        return node.__class__(
+            lhs=qualify_expression(node.lhs),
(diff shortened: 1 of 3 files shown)


coleifer:
I merged a similar patch that is roughly equivalent: 2018e579ac81db5113a26e55ce9078cc80bab446

I have also removed all references to QualifiedNames in the tests -- the helper is no longer needed as the query-builder handles it correctly now.

PR #1780, opened 2018-11-07
fix InterfaceError when postgresql connection is closed

velom:
when `self._state.closed is False` and `self._state.conn.closed != 0`

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -3426,6 +3426,12 @@ class PostgresqlDatabase(Database):
     def get_noop_select(self, ctx):
         return ctx.sql(Select().columns(SQL('0')).where(SQL('false')))
 
+    def is_closed(self):
+        conn_is_closed = not (self._state.conn is not None and self._state.conn.closed == 0)
+        if not self._state.closed and conn_is_closed:
+            self._state.reset()
+        return super(PostgresqlDatabase, self).is_closed()
+
 
 class MySQLDatabase(Database):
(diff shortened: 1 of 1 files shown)


coleifer:
I'm unclear on why this patch is necessary. Granted, Peewee could do a lot more introspection to detect whether the connection is closed or not, but this method is intended more to refer to the "peewee" connection state rather than the "driver" connection state.

coleifer:
I'm going to pass on this -- connections dying unexpectedly probably *should* raise an `InterfaceError`.

velom:
I have described my situation at https://github.com/coleifer/peewee/pull/1779#issuecomment-436895926.
During some exceptions, `self._state.conn.closed` became non zero, but `self._state.closed` is still `False`.

velom:
> I'm going to pass on this -- connections dying unexpectedly probably _should_ raise an `InterfaceError`.

By the way, when connection is dying, peewee doesn't switch `self._state.closed` to `True`.
So every next request will fail here
```python
def cursor(self, commit=None):
        if self.is_closed():
            self.connect()
        return self._state.conn.cursor()
```
because `self.is_closed()` returns `False`, but connection is dead (`self._state.conn.closed != 0`)

coleifer:
Correct - it is necessary to implement reconnect logic.

velom:
do you think, there is no bug in peewee in this case?

PR #1803, opened 2018-12-05
Support order by in aggregate function

lephuongbg:
PostgreSQL support ORDER BY in aggregate function (e.g `json_agg`, `array_agg`, etc.) like this:

```SQL
SELECT array_agg(
    "t1"."name" ORDER BY "t1"."name" ASC, "t1"."created_at" DESC
) AS "users" 
FROM "users" AS "t1"
```

This PR extend `fn` API to support this kind of query.

```
fn.array_agg(User.name).order_by(User.name.asc(), User.created_at.desc())
```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1351,6 +1351,7 @@ class Function(ColumnBase):
     def __init__(self, name, arguments, coerce=True, python_value=None):
         self.name = name
         self.arguments = arguments
+        self._order_by = None
         self._filter = None
         self._python_value = python_value
         if name and name.lower() in ('sum', 'count', 'cast'):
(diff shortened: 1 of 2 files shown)


coleifer:
My own sense is that this is a niche usage and not widely supported.

Rather than implement a special-case for this, I'll suggest using `NodeList` and `SQL` to compose the appropriate "ORDER BY" clause.

coleifer:
```sql
SELECT array_agg(
    "t1"."name" ORDER BY "t1"."name" ASC, "t1"."created_at" DESC
) AS "users" 
```

```python
name_asc = NodeList((User.name, SQL('ORDER BY'), User.name.asc()))
...select(fn.array_agg(name_asc, User.created_at.desc()).alias('users'))
```

PR #1813, opened 2018-12-13
setdefault added to playhouse kv, for dict like setdefault behavior

ascourge21:
To make KeyValue behave more like the regular python dict, a setdefault is also added.
This is a feature that we regularly use in our team when the db isn't initialized with any value, especially for settings and config items.


coleifer:
Thnx.

PR #1843, opened 2019-02-04
Implement connection_connection for Proxy, fixes #1842

### LINKED ISSUE #1842: Cannot use Proxy connection_context decorator.
Defining methods or functions with the `connection_context` decorator does not work with Proxies.

It throws `AttributeError: Cannot use uninitialized Proxy`.

Although I see no reason, why this should be necessary at definition time.

```python
from peewee import Proxy, Model, TextField

database = Proxy()

class User(Model):
    username = TextField(unique=True)

    class Meta:
        database = database

@database.connection_context()
def get_all_users():
   return list(User.select())

```

MaxNoe:
A fix for #1842, also adding a test case

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -422,6 +422,9 @@ class Proxy(object):
     __enter__ = passthrough('__enter__')
     __exit__ = passthrough('__exit__')
 
+    def connection_context(self):
+        return ConnectionContext(self)
+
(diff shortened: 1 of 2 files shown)


coleifer:
Implementation in 91f18c6, as described in #1842.

PR #1869, opened 2019-02-24
Travis CI recommends removing the sudo tag

cclauss:
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo: false__ in your __.travis.yml__, we recommend removing that configuration_"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ matrix:
         - ./.travis_deps.sh
         - sudo ldconfig
       script: "python runtests.py -v2"
+  allow_failures:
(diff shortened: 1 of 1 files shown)


coleifer:
So this change completely fucks the 3.8 tests and the solution is to mark it as "allow failures"? Nope.

coleifer:
OK, I owe you an apology actually, as the psycopg2 driver installation seems to be failing on 3.8 regardless of any other changes. Sorry about that.

PR #1878, opened 2019-03-09
Update peewee.py

jieshukai:
support mysql comment


coleifer:
I'll pass. Plus you've got tests failing due to unbound fields not having a `model` attribute.

PR #1894, opened 2019-03-25
Add support for Postgres's plainto_tsquery

kkinder:
peewee's `TSVectorField` helpfully includes a `match` method to add a `tsquery` to a query using `to_tsquery`. However, `to_tsquery` is one of two ways you can convert a string to a `tsquery`, per the [Postgres docs](https://www.postgresql.org/docs/9.1/textsearch-controls.html):

> PostgreSQL provides the functions to_tsquery and plainto_tsquery for converting a query to the tsquery data type. to_tsquery offers access to more features than plainto_tsquery, but is less forgiving about its input.

This pull request adds a `use_plain` parameter to `TSVectorField.match`, thus letting the programmer choose whether to use `to_tsquery` or `plainto_tsquery`. That's especially helpful when you're writing an application with casual user input for a search query.

### DIFF
--- a/docs/peewee/playhouse.rst
+++ b/docs/peewee/playhouse.rst
@@ -1042,6 +1042,14 @@ inserting or updating the ``search_content`` field:
 
 .. note:: If you are using the :py:class:`TSVectorField`, it will automatically be created with a GIN index.
 
+To search a ``tsvector`` field, you can use its match method:
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


coleifer:
I've made a slightly more compact patch for the actual implementation, plus some improvements to the docs and added a couple tests:

fa8fc729f37b1ab4da420fb51f8bbc90d4074fb7

PR #1900, opened 2019-03-31
Make sure pwiz take care of camelCase

eggachecat:
Add feature/testcases so that if original table/column name is camelCase the autogenerated one by pwiz will be snake_case

Let's say we have a table named `CamelCaseTableName` who has a column named `camelCase`, with `pwiz`,
## Before
```python
class Camelcasetablename(BaseModel):
    camelcase = CharField(column_name='camelCase')

    class Meta:
        table_name = 'camelCaseTableName'
```

## After
```python
class CamelCaseTableName(BaseModel):
    camel_case = CharField(column_name='camelCase')

    class Meta:
        table_name = 'camelCaseTableName'
```

### DIFF
--- a/playhouse/reflection.py
+++ b/playhouse/reflection.py
@@ -455,6 +455,12 @@ class Introspector(object):
     def __repr__(self):
         return '<Introspector: %s>' % self.metadata.database
 
+    @staticmethod
+    def convert_camel_to_snake(name):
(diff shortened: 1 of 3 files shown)


coleifer:
I appreciate the patch. I've chosen to implement it slightly differently (default to using snake case, more tests, factored table-naming into a function to use in both places it is used).

PR #1947, opened 2019-06-10
Support raising custom DoesNotExist exceptions in models

onyb:
### Why do we need this?

- The current way of raising a custom `DoesNotExist` exception is to override model methods like `.get_by_id()`, catch `DoesNotExist` and re-raise with a custom exception class. This is too much boilerplate for something that should be doable with something like:

    ```py
    class BaseModel(Model):
        DoesNotExist = MyCustomException
    ```
- Django ORM supports it, and so do most other ORMs.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -5857,9 +5857,16 @@ class ModelBase(type):
             setattr(cls, '__repr__', lambda self: '<%s: %s>' % (
                 cls.__name__, self.__str__()))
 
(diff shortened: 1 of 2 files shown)


onyb:
@coleifer Could you please explain the reason why you closed this PR? Is raising custom `DoesNotExist` exceptions already possible with PeeWee, or perhaps there is something you'd like me to improve?

Issue #1963, opened 2019-07-03
Missing on= inference for joins (ValueError: More than one foreign key between.... Please specify which you are joining on.)


nad2000:
The inference of **on=...** has changed in the case with multiple foreign keys to the same table, for example, 

```python
class User(Model):
    ...

class Record(Model):
   user = ForeignKeyField(User)
   created_by = ForeignKeyField(User)
   updated_by = ForeignKeyField(User)
```

an attempt to join `Record` to `User` without given **on=...** (`python Record.select().join(User)`)  fails with:
```
ValueError: More than one foreign key between <Model: Record> and <Model: User>. Please specify which you are joining on.
```

I reckon it used to pick up, by default, the foreign key with the name that matches the calls name, namely, **user** in this example.

coleifer:
My sense is to keep this explicit when multiple FKs are present, regardless of what the previous (2.x?) behavior may have been.

nad2000:
that's a bummer. I have heaps of queries that used to work with 2.x but now are broken (all tables have these created_by and modified_by)

coleifer:
Well let me reconsider.

coleifer:
Merged equivalent patch thanks.

PR #1964, opened 2019-07-03
infer foreign key from the name of the model

nad2000:
fix for #1963


coleifer:
Going to pass on this and require the condition when multiple FKs are present.

Issue #1883, opened 2019-03-13
Playhouse PooledPostgresqlDatabase


cypmaster14:
Hello,

Some unexpected error occurs when the pooled connection is used for a Postgresql database.
```
File "D:\Python36_x64\test_project\lib\site-packages\playhouse\pool.py", line 125, in _connect
    ts, conn = heapq.heappop(self._connections)
TypeError: '<' not supported between instances of 'psycopg2.extensions.connection' and 'psycopg2.extensions.connection'
```

cypmaster14:
Peewee ```3.9.2``` 
Psycopg2 ```2.7.7```

coleifer:
Are you messing with `._connections` in any way? The `._connections` list is a heap of `(timestamp, conn)` 2-tuples.

Perhaps the issue is that there are two connections in the `_connections` heap with the exact same timestamp (down to the microsecond?).

Can you replicate this?

cypmaster14:
I just followed the steps from the docs( ```http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#connection-pool```)

This is the full traceback.
```
[2019-03-13 19:36:23,065] - [CP Server Thread-19] - [ERROR - app.py] Exception on /api/v1/notifications [POST]
Traceback (most recent call last):
  File "D:\Python36_x64\test_project\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "D:\Python36_x64\test_project\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "D:\Python36_x64\test_project\lib\site-packages\flask\app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "D:\Python36_x64\test_project\lib\site-packages\flask\_compat.py", line 35, in reraise
    raise value
  File "D:\Python36_x64\test_project\lib\site-packages\flask\app.py", line 1811, in full_dispatch_request
    rv = self.preprocess_request()
  File "D:\Python36_x64\test_project\lib\site-packages\flask\app.py", line 2087, in preprocess_request
    rv = func()
  File "D:\Projects\test_project\test_api\server.py", line 135, in _db_connect
    test_api.libs.models.database.connect()
  File "D:\Python36_x64\test_project\lib\site-packages\playhouse\pool.py", line 108, in connect
    return super(PooledDatabase, self).connect(reuse_if_open)
[...]

PR #1976, opened 2019-07-24
Avoid situations where connections have same ts.

itsmeczaharia:
Hello,
I was able to reproduce the bug from this issue: https://github.com/coleifer/peewee/issues/1883
Adding a small random number to ts, avoids that race condition.

### DIFF
--- a/playhouse/pool.py
+++ b/playhouse/pool.py
@@ -33,6 +33,7 @@ That's it!
 """
 import heapq
 import logging
+import random
 import time
 from collections import namedtuple
 from itertools import chain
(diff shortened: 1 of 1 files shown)


coleifer:
What's odd is that `time.time()` has microsecond precision on my machine. Granted, calling `time.time()` appears to take only ~80ns, but still there's a lot of stuff going on that would seem to make identical timestamps an impossibility. I'm fine with merging, but I can't wrap my head around how this could happen in the first place.

coleifer:
Fixed and merged an equivalent (but not broken) patch.

PR #2002, opened 2019-08-30
Fix bug with multiple model prefetches

rwrigley:
I found when calling prefetch with a model multiple times but on different rel models would result in only one of the rel models getting the instances populated and all others with empty lists. This changes seems to fix the issue.


coleifer:
Thanks

PR #2018, opened 2019-09-19
Update peewee.py

lvhaidong:
Add field comment info


coleifer:
Nope. That's a mysql-only thing.

lvhaidong:
How to set comment in the model, can be described in the database can see the details

coleifer:
It is not supported by sqlite or postgres.

lvhaidong:
I am use mysql, which field can be described in the database can see the details

geonyoro:
Are you averse to adding dv-specific features?

coleifer:
> Are you averse to adding dv-specific features?

Not at all, but they have to be added in such a way that they are compatible. This patch would break hard if a user added help_text on sqlite or postgres. That's not compatible.

PR #2030, opened 2019-10-03
Use the pythonic way to filter with list comprehensions or generator expressions

eltonvs:
This is a really small refactor in order to use the pythonic way to filter in python.

References:
- https://www.oreilly.com/library/view/python-cookbook/0596001673/ch01s11.html
- https://www.artima.com/weblogs/viewpost.jsp?thread=98196


coleifer:
Gonna pass, thanks though.

PR #2033, opened 2019-10-09
Add ability to update JSON fields that are lists in Model.bulk_update

KarolTrojanowskiDL:
I noticed that using `Model.bulk_update` won't work with `JSONField`s that are top-level lists because by default `Value` will unpack the list into multiple values. Top-level lists are valid JSON objects.

Example (before):
```python
class Item(Model):
    json = JSONField()

item = Item.create(json=['a','b'])
item.json = ['c','d']

Item.bulk_update([item], fields=[Item.json])
```
will raise `OperationalError` because of a malformed update query.

Example (after):
```python
class Item(Model):
    json = JSONField()

item = Item.create(json=['a','b'])
item.json = ['c','d']

Item.bulk_update([item], fields=[Item.json], unpack_values=False)
```
will update the row fine.

This applies to both SQLite and PostgreSQL `JSONField`s.

### DIFF
--- a/docs/peewee/api.rst
+++ b/docs/peewee/api.rst
@@ -4184,13 +4184,17 @@ Model
               :py:meth:`Database.atomic`. Otherwise an error in a batch mid-way
               through could leave the database in an inconsistent state.
 
-    .. py:classmethod:: bulk_update(model_list, fields[, batch_size=None])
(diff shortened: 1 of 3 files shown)


coleifer:
I've patched this in a slightly different way. Am hoping to extend the code elsewhere to use the new `to_value()` interface (as opposed to manually creating Value instances w/the appropriate converter). Thx for the patch, at any rate.

PR #2116, opened 2020-02-28
add jsonb array oid

james-lawrence:
attempts to add JSONB array column to postgresql.

however unsure of how to use it. the generated code looks correct

however these fail:
```python
Model.update(array=[{}]).where(Model.id == 1).execute()
Model.create(array=[{}])
```


coleifer:
I don't think using arrays with jsonb is a sensible thing for anyone to do, since json has a native list type already. Going to pass on this.

james-lawrence:
whether you think it is sensible is immaterial, and postgresql allows it.

Issue #2124, opened 2020-03-04
Peewee incorrectly assumes that a column of the form IntegerField(primary_key=True) is an AutoField()


ilyakamens:
- Peewee 3.13.1
- PostgreSQL 9.6.15
- psycopg2 2.8.2

```
class TestModel(db.Model):
    id = IntegerField(primary_key=True)
```

```
>>> TestModel.create_table()
>>> models = generate_models(db.database)
>>> models["testmodel"]._meta.columns["id"]
<AutoField: testmodel.id>
```

```
# \d testmodel;
             Table "public.testmodel"
 Column |  Type   | Collation | Nullable | Default
--------+---------+-----------+----------+---------
 id     | integer |           | not null |
Indexes:
    "testmodel_pkey" PRIMARY KEY, btree (id)
```

This may be the wrong spot, but it looks like Peewee only checks for primary key + IntegerField here: https://github.com/coleifer/peewee/blob/master/playhouse/reflection.py#L166-L179

coleifer:
Yeah, Peewee uses a heuristic when introspecting the database and assigns the auto-incrementing integer field when it discovers integer primary keys. Known behavior.

ilyakamens:
@coleifer - Understood. Would you accept a PR to fix this case?

coleifer:
@ilyakamens - before putting together a patch, how would you go about handling this? My thought was that by far the most common case of using an integer primary key was as an auto-incrementing id. On the other hand, determining if an integer primary-key actually *is* autoincrementing in a cross-db compatible way (without introducing false-negatives, which would be much worse IMO than false-positives), is kinda tricky.

It's been a while since I looked into it, though, so maybe you have some ideas? The main things to keep in mind, as far as I'm concerned, are:

* peewee's reflection has worked this way since the beginning and this is the first issue about this
* it seems much easier to deal with a false-positive than a false-negative. Peewee let's you manually specify the primary key in either case, so i'm kinda pressed to think of when this would even negatively impact someone.
* is there an extremely reliable way to detect autoinc primary keys across the supported dbs? Or conversely, to detect when an integer primary key is definitely not autoinc?

ilyakamens:
I started working on it already before your reply. 🙂 

> My thought was that by far the most common case of using an integer primary key was as an auto-incrementing id.

[...]

PR #2126, opened 2020-03-05
Improve reflection for AutoField/IntegerField

ilyakamens:
Specifically, don't assume `IntegerField(primary_key=True)` is an `AutoField()` if the database isn't SQLite.

See https://github.com/coleifer/peewee/issues/2124 for more context.


coleifer:
I'm going to pass on this for now, but may reconsider it at some point.

PR #2132, opened 2020-03-13
ensure extracting Field instance from Function arguments

kamilkozik7:
when the Function is nested in another Function in _initialize_columns method of class BaseModelCursorWrapper


coleifer:
I'm going to pass. If you have complicated nesting and need a special conversion, just specify a `python_value()` converter to the `fn.(...)`.

PR #2190, opened 2020-05-27
fix diary example

phanirithvij:


phanirithvij:
wow that was fast.

phanirithvij:
The build failed for some reason might want to revert?

coleifer:
https://travis-ci.org/github/coleifer/peewee/builds/691771699

Issue #1835, opened 2019-01-26
Model.get() method fails silently


arel:
I came across a behavior with [`Model.get`](http://docs.peewee-orm.com/en/latest/peewee/api.html#Model.get) I found confusing and could be improved or at least better documented.

Calling `Model.get()` is a shorthand for "selecting with a limit of 1", which is clear. When you provide it filters, such as `.get(id=1442)`, things work as expected too. What was unexpected to me was that if you forget to specify the filter parameter by name such as `.get(1442)` the query does *not* fail. Rather, it returns an arbitrary result as if `.get(True)` or simply `.get()` were given. In my case, this led to a subtle bug because the two results were coincidentally very similar. For instance:

```
In [15]: Image.get(id=1442)                                                                                                                        
Out[15]: <Image: 1442>

In [16]: Image.get(1442)                                                                                                                           
Out[16]: <Image: 1433>
```

Is this the intended / desired behavior for some reason I'm missing? Or, would it be possible to raise an exception whenever the query expression is not an instance of type [Expression](http://docs.peewee-orm.com/en/latest/peewee/api.html#Expression)?

P.S. thanks for your work on this library!

coleifer:
This gets translated into:

```sql
select * from image where 1442
```

So the db interprets that, most likely, as simply `true` and returns the first row of this arbitrarily-large result set.

I'm aware of this issue, but not positive about the best way to fix it. Python is highly dynamic, and so is peewee. Adding validation to all SQL clauses (not just WHERE, which is in this example), seems prohibitively complicated.

Perhaps, though, the `.get()` method can be fixed up.

coleifer:
I've chosen to resolve this by supporting .get() with a single integer parameter, treating it as `.get(primary-key == value)`. So in the common case where one just passes an int, it will work as expected.

arel:
OK, thanks!

PR #2232, opened 2020-07-29
Support queries of all types of pkey in `get()`

wooparadog:
In https://github.com/coleifer/peewee/issues/1835 get by primary key is added for `int`, but there're cases when primary key is other than integer, especially in analytics domain.

Using `Node` to exclude `Expression` seems a good idea, need more insight though.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -6395,7 +6395,7 @@ class Model(with_metaclass(ModelBase, Node)):
         sq = cls.select()
         if query:
             # Handle simple lookup using just the primary key.
-            if len(query) == 1 and isinstance(query[0], int):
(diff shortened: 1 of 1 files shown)


peewee.py
```diff
@@ -6395,7 +6395,7 @@ def get(cls, *query, **filters):
         sq = cls.select()
         if query:
             # Handle simple lookup using just the primary key.
-            if len(query) == 1 and isinstance(query[0], int):
+            if len(query) == 1 and not isinstance(query[0], Node):
```
wooparadog, line 6398: Maybe I should add `Context` here as well?

coleifer:
I'm going to pass, but thanks.

PR #2236, opened 2020-07-31
Added JSONField

daveusa31:


coleifer:
I'm going to pass. There are vendor-specific JSON field implementations, which are recommended. Those who wish to just store json in a textfield can subclass the existing TextField implementation or use a 3rd party package.

PR #2267, opened 2020-09-11
Update SQLite max_variable_number info

jonashaag:
Missed in #2263


coleifer:
Thanks merged an equivalent patch w/slightly different wording.

PR #2278, opened 2020-10-07
GeneratedHashField

milesgranger:
Hi! :wave: 

First, thank you so much for such a great library. :+1: 

Feature request :ribbon: 
---
I would like to make use of PostgreSQL's `GENERATED ALWAYS AS` statement, for computing a hashed key based on one or more other fields. Specifically for a generating a hash

Solution :question: 
---
I've provided a prototype which accomplishes this, but it could use some updates for generalization (maybe not hash, different types, etc). Do you think this is worth pursuing or maybe best left as a custom implementation for private use? I suppose it is similar to SqlAlchemy's [Computed](https://docs.sqlalchemy.org/en/13/core/defaults.html?highlight=server#computed-generated-always-as-columns)

### DIFF
--- a/playhouse/postgres_ext.py
+++ b/playhouse/postgres_ext.py
@@ -491,3 +491,17 @@ class PostgresqlExtDatabase(PostgresqlDatabase):
         if named_cursor:
             cursor = FetchManyCursor(cursor, array_size)
         return cursor
+
+
+class GeneratedHashField(CharField):
+    def __init__(self, *key_cols, **kwargs):
+        self._key_cols = key_cols
+        super().__init__(**kwargs)
+
+    def ddl(self, ctx):
+        node_list = super().ddl(ctx)
+        columns = " || '-' || ".join(f"coalesce({key}, '')" for key in self._key_cols)
+        sql = SQL(
+            f"GENERATED ALWAYS AS (encode(sha256(({columns})::bytea), 'hex')) STORED"
+        )
+        return NodeList((node_list, sql))


conqp:
F-Strings are not supported in Python 2.x and Python 3 < 3.6.
This is why the check fails.

milesgranger:
I see that, thanks! :+1:  Didn't think it was worth fixing right away since I imagine changes will be needed anyway, if this is at all even wanted.

coleifer:
I'm not interested in maintaining or supporting this, but feel free to offer it as a 3rd-party extension.

milesgranger:
Thank you for your input.

I don't see anything similar to [Computed](https://docs.sqlalchemy.org/en/13/core/defaults.html?highlight=server#computed-generated-always-as-columns) offered by Peewee (I couldn't locate anything in the docs), or is this the right approach?

coleifer:
No but since Postgres and Sqlite have the same syntax it's probably worth adding. I will look into this.

coleifer:
As I was implementing this, I ran into a number of subtle differences between the Sqlite, MySQL, MariaDB and Postgres implementations. For now, I think users will be best-off using the existing `constraints=` parameter to specify a generated column definition, although the code is in `feature/computed` branch if you want to fork it.

```python

constraint = SQL("generated always as (first || ' ' || last)")
full_name = CharField(constraints=[constraint])
```

milesgranger:
Lovely! Thanks for the tip, I really appreciate it.

PR #2279, opened 2020-10-07
Enable use of range for IN filters in python 3

conqp:
While it is currently possible to do something like

`Model.select().where(Model.id << range(42, 256))`

in Python 2, this is not possible in Python3.
The reason being, that `range()` returns a `list` in Python 2, which is currently a member of `Value._multi_types`, but is of type `range` in Python 3.
Therefor I used the existing Python 2/3 branch logic to declare `multi_types` globally with other version-specific constants and used it in `Value` instead.

I did not yet add a test case for this, since `test_select_in_list_of_values` in `sql.py` currently only handles strings.


coleifer:
I've merged an equivalent patch. I had hoped to also support `map` but that's more like a generator and doesn't support multiple iterations, so that doesn't work. At any rate, thanks for the suggestion.

PR #2284, opened 2020-10-22
Use link_shared_object instead of link_executable in SQLite detection

sboisson:
`link_executable` doesn't use environment variable `LDFLAGS`, only environment variable `CC` (which usually only contains the name of the compiler, not the linking options)… which can be problematic and cause false negative on setups where SQLite is not installed in system path, but on a custom path.

Also the setup is not attempting to build an executable, but shared objects.


coleifer:
Excellent, thank you!

PR #2288, opened 2020-10-27
Fix typo in querying docs

pylipp:


coleifer:
merged an equivalent patch

PR #2305, opened 2020-11-26
Unexpected column names when using fn and .dicts

albatros69:
Consider the following example (I tried to shave it to a minimum):
```python
class Test(Model):
    start_date = DateTimeField(null=True)
    end_date = DateTimeField(null=True)

>>> list(Test.select(fn.MIN(Test.start_date), fn.MAX(Test.end_date)).dicts())
[{'start_date")': datetime.datetime(2020, 9, 11, 9, 56), 'end_date")': datetime.datetime(2020, 11, 29, 16, 0)}]
```

See the extra quote and parenthesis for the column names. One would expect something more like:
```python
>>> list(Test.select(fn.MIN(Test.start_date), fn.MAX(Test.end_date)).dicts())
[{'start_date': datetime.datetime(2020, 9, 11, 9, 56), 'end_date': datetime.datetime(2020, 11, 29, 16, 0)}]
```
I'm just proposing here to strip the colum names from the extra right parenthesis.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -7378,7 +7378,7 @@ class BaseModelCursorWrapper(DictCursorWrapper):
             if dot_index != -1:
                 column = column[dot_index + 1:]
 
-            column = column.strip('"')
+            column = column.strip('")')
             self.columns.append(column)
             try:
                 raw_node = self.select[idx]


coleifer:
I'm OK with adding this, but you should *really* be explicitly aliasing these so that it is not database-dependent and fragile. I believe, furthermore, this example is specific to sqlite - postgres behaves differently (I believe the columns will be named "min" and "max", at least that's how it behaves on my laptop with postgres 13), and older versions of sqlite behave even differently if I recall correctly.

I've implemented a similar patch which also adds a test-case and merged.

albatros69:
Thanks a lot for the answer. You're absolutely right about the aliasing necessity. I discovered this issue as part of some quick&dirty debugging, where i didn't take the time to alias (which I usually do). As it was quite straight forward to correct, I thought it was better to consolidate that in a PR.

PR #2329, opened 2021-01-22
Allow deferred foreign keys on lazy loading.

### LINKED ISSUE #2328: Allow deferring of foreign keys when lazy_load is disabled
I just encountered a problem when disabling lazy loading on foreign keys.

```python
class Referenced(Model): pass

class Referencing(Model):
     ref = ForeignKeyField(Referenced, lazy_load=False)

referenced = Referenced()
referencing = Referencing()
referenging.ref = referenced
referenced.save()
referencing.save()
```
The above code will result in a `DoesNotExist` exception, because the current `NoQueryForeignKeyAccessor` does not attempt to check `instance.__rel__`
I think this should work, since it does not require a query.
I will make a pull request, that solves this issue and also gets rid of the `NoQueryForeignKeyAccessor`.
This has the beautiful side effect, that changing a `ForeginKeyField`'s `lazy_load` attribute during runtime actually has an effect on the behaviour. Currently, `lazy_load`  is only evaluated during `ForeignKeyField.__init__` and then unnecessarily stored on the instance, since subsequent changes to it will not have any effect.

conqp:
Fixes #2328.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4393,10 +4393,10 @@ class ForeignKeyAccessor(FieldAccessor):
     def get_rel_instance(self, instance):
         value = instance.__data__.get(self.name)
         if value is not None or self.name in instance.__rel__:
-            if self.name not in instance.__rel__:
+            if self.name not in instance.__rel__ and self.field.lazy_load:
                 obj = self.rel_model.get(self.field.rel_field == value)
                 instance.__rel__[self.name] = obj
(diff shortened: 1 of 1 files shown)


coleifer:
Thanks for this - I've merged an equivalent patch which includes a regression test as well: b34e60154fbb7e2a0aabd271dd55017f1b0af8f0

PR #2353, opened 2021-02-19
Attempt new connections if the server is not ready

pierky:
When using `ReconnectMixin`, the execution of a SQL statement is attempted again if it fails because of "operational issue" (for example when the connection to the DB server is lost). During the reconnection phase, the DB server might not be ready to handle new connections, thus the call to `self.connect()` could also fail.

When using `ReconnectMixin`, the execution of a SQL statement is attempted again if it fails because of "operational issue" (for example when the connection to the DB server is lost). During the reconnection phase, the DB server might not be ready to handle new connections, thus the call to `self.connect()` could also fail.

With this PR I'm proposing an infinite loop to attempt new connections indefinitely until one is succeeded. Not the perfect solution, I know, I'd just like to see if something along this line would be accepted (I know about some *important* opinions about the disconnection handling). If so, probably some mechanism can be added to the `ReconnectMixin` class to let the developer to specify a limit of attempts or to somehow configure the behaviour.

### DIFF
--- a/playhouse/shortcuts.py
+++ b/playhouse/shortcuts.py
@@ -1,3 +1,5 @@
+import time
+
 from peewee import *
 from peewee import Alias
 from peewee import CompoundSelectQuery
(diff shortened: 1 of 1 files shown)


coleifer:
>Not the perfect solution, I know, I'd just like to see if something along this line would be accepted

I'd suggest putting this into your own subclass of `MySQLDatabase` (or `ReconnectMixin`, whichever you use) as I don't intend to maintain this code or advocate this approach.

PR #2369, opened 2021-03-05
Fix _generate_insert for nullable Fields

### LINKED ISSUE #2368: insert_many() does not generate NULL values for fields with null=True
Snippet with error.
```python
import peewee

db = peewee.SqliteDatabase('people.db')


class Model(peewee.Model):
    name = peewee.CharField()
    x = peewee.IntegerField(null=True)

    class Meta:
        database = db


db.connect()
db.create_tables([Model])
query = Model.insert_many([{"name": "foo", "x": None}, {"name":"bar"}])
query.execute()
```
In a workaround i have to use callable, because of `default=None` is ommited:
```python
class Model(peewee.Model):
    name = peewee.CharField()
    x = peewee.IntegerField(null=True, default=lambda: None)

    class Meta:
        database = db
```

rafal-jaworski:
I don't know if test is in good file and if the solution is suitable :)
Fixes #2368

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2671,6 +2671,8 @@ class Insert(_WriteQuery):
                         val = defaults[column]
                         if callable_(val):
                             val = val()
+                    elif isinstance(column, Field) and column.null:
+                        val = None
                     else:
                         raise ValueError('Missing value for %s.' % column.name)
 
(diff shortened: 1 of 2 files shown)


coleifer:
I've merged an equivalent patch. Thanks!

PR #2382, opened 2021-04-02
Add ModelSelect.get_or_none()

ilyakamens:
Mr. Leifer,

First, thanks for creating and maintaining such a great library. We've used it at Propel for many years now.

This PR adds `get_or_none()` to `ModelSelect` in an effort to make the querying API a little more consistent. `Foo.get_or_none()` is very convenient, and we'd like the same sort of convenience for things like `Foo.select().join(Bar).where(Foo.baz == baz).get_or_none()`.


coleifer:
Thanks for the patch, looks good and merged.

PR #2384, opened 2021-04-03
Fix reference to `first` method

### LINKED ISSUE #2383: `Model.first()` doc reference broken
In the docs under [Selecting a single record](https://github.com/coleifer/peewee/blob/master/docs/peewee/querying.rst#selecting-a-single-record) there is a broken link to `Model.first`. Using it yields an attribute error that `first` doesn't exist on my model. I fixed this by using `Model.select().first()` and most likely the docs also need to reference the `first` method of the `ModelSelect` class instead?

Helveg:
Not sure that this is the right class but it closes #2383


coleifer:
Eh I'll need to look into this some more - not urgent.

PR #2391, opened 2021-04-16
Add ability to create an instance from a back reference

mikemill:
Allow an instance of a model to be created using the back reference.  For example, if I have a `User` and I want to create a `Tweet` for that user then allow it to be used as `user.tweets.create()`.

On a practical level this helps a lot when there are logs associated with a user or similar object.  Compare the above with having to import the log model and do `Log.create(user=user, otherparam=value)` which can cause circular import issues with large projects.

The patch provide is intended to provide an example.  I tried to fit the style but am open to any changes needed.

Thanks

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4445,7 +4445,7 @@ class BackrefAccessor(object):
         if instance is not None:
             dest = self.field.rel_field.name
             return (self.rel_model
-                    .select()
+                    .backref_select(self.field, getattr(instance, dest))
                     .where(self.field == getattr(instance, dest)))
         return self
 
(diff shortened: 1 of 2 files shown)


coleifer:
The patch looks fine, but I have tried to stay away from that particular pandora's box. In the simple case it's straightforward, but it breaks the backref abstraction a bit, and there are just a few too many edge-cases. I prefer to keep it as a naive/simple `Select`.

If you really want this in your own app, I've pushed a patch 626b34856eeb842ad27c885f3c1fba95cce78399 that allows you to override the backref accessor class on your foreign-keys - making it hopefully easier to extend and use your own implementation.

PR #2397, opened 2021-04-30
Added two special expressions, for BooleanField

daveusa31:
To write queries like this,
```python
User.select().where(User.is_activated.is_true())
```
And not like this
```python
User.select().where(User.is_activated == True)
```


coleifer:
You don't even need `.is_true()` honestly, you can just:

```python
.where(User.is_activated)
```

Or for false:

```python
.where(~User.is_activated)
```

This method seems quite redundant.

PR #2431, opened 2021-06-15
feat(style): let the output follows PEP8

HydrogenDeuterium:
when using pwiz generates file to import,IDE always warning 'PEP 8: E302
expected 2 blank lines, found 1'.
I try to fix it.


coleifer:
Eh, I'll pass.

Issue #2444, opened 2021-07-05
IN queries on  fields with iterable Python representations


dustinrb:
When using arrays in a JSON field, the IN operator unpacks all the values into a single list. For example:

```python
itr_select_query = (KeyData
    .select()
    .where(KeyData.data.in_(
        [[1], [2], [1,2]]
    )))
```

will generate the SQL query:
```python
('SELECT "t1"."id", "t1"."key", "t1"."data" FROM "key_data" AS "t1" WHERE ("t1"."data" IN ((json(?)), (json(?)), (json(?), json(?))))', ['1', '2', '1', '2'])
```

instead of:
```python
('SELECT "t1"."id", "t1"."key", "t1"."data" FROM "key_data" AS "t1" WHERE ("t1"."data" IN ((json(?)), (json(?)), (json(?), json(?))))', ['[1]', '[2]', '[1, 2]'])
```

dustinrb:
I see why you closed #2445;  it didn't make sense to single out the IN clause in that part of the code.

I think the broader issue is that you cannot use any field for an IN query where it's Python representation is a sequence. You might be able to get around this by calling

```python
...where(KeyData.data.in_(KeyData.data.db_value(i) in [[1], [2], [1,2]]))...
```
but that seems unintuitive from the user perspective. 

Perhaps the better solution would be to handle this in the `Field._in` with the understanding that if the user provides an iterable that each item has `Field.db_value` called on it before it's formed into an Expression.

I'm happy to work on a patch. I'm just too unfamiliar with codebase to architect this.

dustinrb:
I've implemented a [fix](https://github.com/dustinrb/peewee/tree/json_in_query) that converts `multi_types` to Values in the Expression generated by the `.in_` function instead of mutating one SQL generation.

This process occurs outside of a Context block. Will that pose any issues the node tree when generating SQL?

coleifer:
I looked at the patch and it has the same problem.

For now, you can always try:

```python
vls = [Value(l, unpack=False) for l in list_of_lists]
.where(KeyData.data.in_(vls))
```

dustinrb:
Thanks for looking into this.

Maybe I'm not sure what the exact issue is. Your interim fix is essentially what my second patch does in the Field class which happens before SQL generation. Are you saying that query input should be untouched before it's handed over to the SQL generating code? In that case it feels like a better solution would be to add a flag to allow shallow unpacking for a single Expression and/or NodeList.

[...]

PR #2445, opened 2021-07-05
Fixes bug in JSON IN query for iterables

dustinrb:
Resolves issue #2444


coleifer:
I don't like this patch for a couple of reasons, nor am I sure I understand the goal behind supporting such a query. Namely because it mutates the expression state when sql() is called, which is not done elsewhere in the codebase. Additionally, searching for an exact array within a json field seems problematic because there's no possibility of doing so efficiently from the database perspective.

I'm going to pass on this patch, but will look at the original issue and give it some thought.

PR #2447, opened 2021-07-15
Delete TODO.rst

alexpirine:
Seems like it hasn't been used for two years.


coleifer:
I like to keep it as a placeholder in case I want to fill it in.

PR #2455, opened 2021-08-08
Update api.rst

aparcar:
The example is wrong since it doesn't split the `idlist`.

This is the same function as used in the test. However the example has a case with an empty list (`[]`) which won't work, since `None.split(",")` won't work. Any simple solutions here?


aparcar:
Actually the solution is as simple as `(idlist.split(",") if idlist else [])`

coleifer:
a643a6952372e22ce658c8ca9a23322bb4b2ab57

PR #2483, opened 2021-11-04
Insert link to gevent docs for monkey-patching

pylipp:


pylipp:
Not a fan of links becoming stale?

Issue #1302, opened 2017-06-29
Foreign Key "ID descriptor" name incorrect when using `db_column`


ibushong:
```python
class Event(BaseModel):
    venue = ForeignKeyField(Venue, db_column='place_id')

...

Event.select().where(Event.venue == 5)  # this works fine
Event.select().where(Event.venue_id == 5)  # --> AttributeError: type object 'Event' has no attribute 'venue_id'
Event.select().where(Event.place_id == 5)  # --> This is successful, though, IMO, the above expression should be the correct one.
```

Hmm actually I guess I could see this being "correct". But just wanted to make sure it was expected.

coleifer:
I definitely understand your confusion, and likely the docs aren't clear on this point.

Inside `ForeignKeyField.add_to_class()` you can see that the "id descriptor" gets its attribute from the db_column (if set) and otherwise will be the field's name + "_id":

```python
        self.db_column = obj_id_name = self.db_column or '%s_id' % self.name
        if obj_id_name == self.name:
            obj_id_name += '_id'
```

So what you're seeing is correct according to the above logic.

ibushong:
Yeah, I eventually found that. I guess I just felt that if you're "renaming" a foreign key field, then both the name-descriptor and id-descriptor values would change accordingly. My goal in the example was to entirely replace the usage of place/place_id with venue/venue_id (without having to rename the actual database column). When I'm thinking about the venue FK, I want to hide/abstract-away the fact that under the hood it's actually referencing the `place_id` column.

coleifer:
I don't want to break backwards compatibility in case anyone was relying on the behavior described in your issue, *however* I've added a new parameter to `ForeignKeyField` that allows you to specify the name of the "object id descriptor". Your declaration would read:

```python
venue = ForeignKeyField(Venue, db_column='place_id', object_id_name='venue_id')
```

ibushong:
Awesome. That's a good solution. Thanks!

PR #2485, opened 2021-11-11
Document example usage of object_id_name

pylipp:
In combination with `column_name` defined, an AttributeError is raised otherwise.
This behavior was confusing me at first, until I found #1302. The docs (apart from the ForeignKeyField API) don't mention object_id_name


coleifer:
I found the text and admonition confusing, even in context. I'll pass for now but will think about whether some improvement to the docs is needed for this situation.

PR #2499, opened 2021-12-08
Add BlackSheep integration

q0w:


coleifer:
I really can't keep up with these new async-framework-of-the-day things. Additionally, connection handling doesn't work the way you might expect with async, so this is probably incorrect anyways.

Issue #1289, opened 2017-06-06
Mysql "USE INDEX" clause?


ParthGandhi:
Is there a way to use this option? Or in general, to add arbitrary clauses to the query?

Currently i've hacked the internal `QueryCompiler` to allow adding the required clause, but thats dirty as hell cause it forces me to copy the entire method into the app code and add only 3 lines to it.

Here it is:
```python
    def generate_select(self, query, alias_map=None):
        from peewee import CompoundSelect, _StripParens, SQL, EnclosedClause, Clause, CommaClause

        model = query.model_class
        db = model._meta.database

        alias_map = self.calculate_alias_map(query, alias_map)

        if isinstance(query, CompoundSelect):
            clauses = [_StripParens(query)]
        else:
            if not query._distinct:
                clauses = [SQL('SELECT')]
            else:
                clauses = [SQL('SELECT DISTINCT')]
                if query._distinct not in (True, False):
                    clauses += [SQL('ON'), EnclosedClause(*query._distinct)]

            select_clause = Clause(*query._select)
            select_clause.glue = ', '

            clauses.extend((select_clause, SQL('FROM')))
            if query._from is None:
                clauses.append(model.as_entity().alias(alias_map[model]))
            else:
                clauses.append(CommaClause(*query._from))

        """CUSTOM CODE START"""
        if query._use_index is not None:
            clauses.append(SQL(f'USE INDEX({query._use_index})'))
        """CUSTOM CODE END"""

        if query._windows is not None:
            clauses.append(SQL('WINDOW'))
            clauses.append(CommaClause(*[
                Clause(
                    SQL(window._alias),
                    SQL('AS'),
                    window.__sql__())
                for window in query._windows]))

        join_clauses = self.generate_joins(query._joins, model, alias_map)
        if join_clauses:
            clauses.extend(join_clauses)

        if query._where is not None:
            clauses.extend([SQL('WHERE'), query._where])

        if query._group_by:
            clauses.extend([SQL('GROUP BY'), CommaClause(*query._group_by)])

        if query._having:
            clauses.extend([SQL('HAVING'), query._having])

        if query._order_by:
            clauses.extend([SQL('ORDER BY'), CommaClause(*query._order_by)])

        if query._limit is not None or (query._offset and db.limit_max):
            limit = query._limit if query._limit is not None else db.limit_max
            clauses.append(SQL('LIMIT %s' % limit))
        if query._offset is not None:
[...]

PR #2501, opened 2021-12-10
Allow to use peewee.SQL from `from_`

D0d0:
**Problem:**
Tried to apply custom SQL to `from_` function from this example https://github.com/coleifer/peewee/issues/1289#issuecomment-307128658 However in my query, I use some joins.

Here is example
```python
q = User.select(User, Tweet).from_(SQL(f'`users` as `t1` USE INDEX(`user_id`) ')).join(Tweet)
for x in q:
   print(x)
```

But without fix, exception is thrown
```
  File "JetBrains\PyCharm2021.2\scratches\scratch_71.py", line 27, in <module>
    for file in q:
  File "venv\lib\site-packages\peewee.py", line 4408, in next
    self.cursor_wrapper.iterate()
  File "venv\lib\site-packages\peewee.py", line 4325, in iterate
    self.initialize()  # Lazy initialization.
  File "venv\lib\site-packages\peewee.py", line 7587, in initialize
    if curr not in self.joins:
TypeError: unhashable type: 'SQL'
```

**Solution**
When initializing ModelCursorWrapper skip SQL instances

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -7596,6 +7596,9 @@ class ModelCursorWrapper(BaseModelCursorWrapper):
                 accum.append(curr.rhs)
                 continue
 
+            if isinstance(curr, SQL):
+                continue
+
             if curr not in self.joins:
                 continue
 


coleifer:
I don't really like this solution. The join will not correctly set-up the model graph, so the best solution in my opinion is to write:

```python
q = User.select(User, Tweet).from_(SQL(f'`users` as `t1` USE INDEX(`user_id`) ')).join(Tweet)
for x in q.objects():
   # assume username is on User table, and content is on Tweet table.
   print(x.username, x.content)
```

PR #2521, opened 2022-01-30
Added optional use_column_names instead of field names in model_to_dict

deduble:
I am using this in model_to_dict in jobs because it makes the created python object keys consistent with the database column names when communicating through web apps. Despite column names are not case sensitive when querying, language mediums are, so it goes hand in hand with pwiz db_model.py tool. I didn't know what to do with backrefs though, hence left those lines.


coleifer:
I'll pass on this, thanks.

PR #2576, opened 2022-06-07
Correct Model.select retreiving method in docs

sofienekhiari:
I believe there's a mistake in here :)


coleifer:
Wrong. It's correct.

PR #2585, opened 2022-07-03
Update database.rst

inktrap:
If the apsw **reference is rendered,** the text reads (emphasis mine):

> Peewee also comes with an alternate SQLite database that uses **apsw, an advanced sqlite driver**, an advanced Python SQLite driver.

I know that the idea is to describe the reference without relying on the self-description of the reference, but in this case they are (almost)identical. Another way would be to shorten the ref to just apsw, if restructured text permits that.


coleifer:
Thanks, I merged a roughly equivalent patch.

PR #2597, opened 2022-07-21
Modify the incorrect description of the document

sunmd01:


coleifer:
I'll see about rewording this later. The docs are trying to show what the available options are, and their default values.

PR #2603, opened 2022-08-07
Added special expressions to boolean field

alteralt:
Added an expression so that in the request instead of:
```python
Model.select().where(Model.boolean_field == True)
````
To write:
```python
Model.select().where(Model.boolean_field.is_true())
````

The same expression already exists:
https://github.com/alteralt/peewee/blob/b8c629bc100c4886c5b87b6eb62cce66c84a0d9c/peewee.py#L1197

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -5253,6 +5253,15 @@ class BooleanField(Field):
     field_type = 'BOOL'
     adapt = bool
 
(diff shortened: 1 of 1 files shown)


coleifer:
Going to pass for now, thanks.

PR #2611, opened 2022-08-26
Add `SelectBase.scalars()`

ilyakamens:
Mr. Leifer, thanks for creating and maintaining such a great library. We've used peewee at Propel since 2015.

This PR adds `SelectBase.scalars()`, which is something we've wanted for a while. `scalar()` is very convenient, and we thought something similar for multiple values would be too. I've proposed `scalars()` as the name, but would be happy to settle on something else if you prefer.

Thanks for considering.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2212,6 +2212,10 @@ class SelectBase(_HashableSource, Source, SelectQuery):
         return row[0] if row and not as_tuple else row
 
     @database_required
+    def scalars(self, database):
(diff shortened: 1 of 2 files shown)


coleifer:
There are a few issues with your patch - it consumes the entire cursor in one go and returns a list, when it would probably be better suited to use a generator. Similarly, it is not carrying through the `database` parameter. I will implement a similar patch and merge it.

ilyakamens:
Thank you!

Issue #2631, opened 2022-10-18
Support LIMIT with prefetch and MySQL


nomme:
Hi

The premise is simple. It would be pretty sweet if peewee's `prefetch` supports `LIMIT` when using MySQL. I want to add support for it by introducing a switch on `prefetch` that subqueries using `INNER JOIN` instead of `IN`.

Background:
In a project using peewee and MySQL we often encounter the following limitation when using `prefetch`.
```
This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'
```

This can be seen reported multiple times across the internet, which you also address all those times. [There are ways around it](https://github.com/coleifer/peewee/pull/381#issuecomment-47823210) but native support would feel much better, and would also require one less query and avoid giant `IN` lists with large limit values.

I'll submit a PR in a moment to show what I have in mind.

coleifer:
Merged your PR here: af3f34f10555a22003726dca90fb2d577617656a

nomme:
Awesome, thank you!

PR #2632, opened 2022-10-18
Support IN and INNER JOIN subqueries on prefetch

nomme:
This PR intends to solve #2631 

It adds a parameter to `prefetch` that allows the user to choose the type of subquery performed. Using `INNER JOIN` is desirable when combining `LIMIT` with MySQL as the current implementation is not compatible.

Usage example 
```python
from peewee import PREFETCH_TYPE

people = Person.select().order_by(Person.name)
query = people.prefetch(Note, NoteItem, prefetch_type=PREFETCH_TYPE.JOIN)
```

The current behavior should remain the same as `prefetch_type` default to using the current implementation. I suggest whitespace changes are ignored when diffing `prefetch_tests.py`.

It goes without saying, but any feedback is welcome.

### DIFF
--- a/docs/peewee/api.rst
+++ b/docs/peewee/api.rst
@@ -4784,13 +4784,19 @@ Model
 
         Use Django-style filters to express a WHERE clause.
 
-    .. py:method:: prefetch(*subqueries)
+    .. py:method:: prefetch(*subqueries[, prefetch_type=PREFETCH_TYPE.WHERE])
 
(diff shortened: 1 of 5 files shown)


coleifer:
Nice patch, thank you. I've rebased your commits into a single commit af3f34f10555a22003726dca90fb2d577617656a and amended it to respect the line-length limit, but otherwise your changes are all merged.

PR #2648, opened 2022-11-12
Remove redundant wheel dep from pyproject.toml

mgorny:
Remove the redundant `wheel` dependency, as it is added by the backend automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools"]


coleifer:
I'm just following what is listed here: https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour

And as recently as this week by another one of the pypa team: https://github.com/pypa/pip/issues/8559#issuecomment-1310317396

mgorny:
> I'm just following what is listed here: https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour

Well, I don't think it's needed for the "fallback" code path either but even if it were, you are not supposed to use that backend. It's just "what happens when there's no pyproject.toml".

> And as recently as this week by another one of the pypa team: [pypa/pip#8559 (comment)](https://github.com/pypa/pip/issues/8559#issuecomment-1310317396)

Unfortunately, it will probably take years before people realize they're copying a mistake from the past. That's why I'm always linking to the official documentation.

coleifer:
>Unfortunately, it will probably take years before people realize they're copying a mistake from the past. That's why I'm always linking to the official documentation.

I agree, and would add my own opinion that pypa is an absolute joke.

I'm going to leave it in. Call it cargo-culting, call it senseless, but I just don't care to spend the effort to try and learn all the intricacies of these various new tools/their interactions/and what combination of what versions does what (wheel, pyproject, pip, setuptools). They'll probably change everything again in a year or two.

mgorny:
Well, the whole point is that they are considering actually stopping to use `wheel` (though it was a while back, back when wheel received zero maintenance) in favor of a built-in wheel generator in setuptools.

PR #2677, opened 2023-02-09
query.first() mutates the query

klen:
The `query.first()` method mutates the query and it may lead to unexpected behaviour

```python
query = SomeModel.select()
first_item = query.first()
for item in query:
    # This cycle only runs once, because after first() the query is mutated and now it has limit 1
```

The PR fixes the behaviour.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -2215,10 +2215,10 @@ class SelectBase(_HashableSource, Source, SelectQuery):
 
     @database_required
     def first(self, database, n=1):
+        query = self
         if self._limit != n:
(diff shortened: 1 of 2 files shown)


coleifer:
The `first()` method is a bit of a special-case. It applies a limit of 1 to avoid making the server generate the full result-set, as it is assumed that after calling `first()` the query itself will not be iterated. Similarly, the mutation is done in-place so that multiple calls to `first()` do not result in additional queries.

I will add some documentation explaining this.

PR #2698, opened 2023-03-26
Ae configurable default

paetling:
## Context
Make it so that we can configure the type of the id that is created


coleifer:
Is there a reason you are not just using:

```python
class Base(Model):
    id = BinaryUUIDField(primary_key=True, ...)

class MyModel(Base):
    # MyModel.id will be a binary uuid field
```

This change seems redundant.

Issue #2707, opened 2023-04-14
 pwiz generates detailed field attributes


GuoGuo54:
The peewee model generated by pwiz based on the database table is too simple， like：

class NameModel(Model):
    name = CharField(null=True)
    status = IntegerField(default=11, index=True)

    class Meta:
        table_name = 'name_model'

But generally we need more detailed field attributes，like (max_length, verbose):
 
class NameModel(Model):
    name = CharField(max_length=128, null=True, verbose_name='name')
    status = IntegerField(default=11, index=True, verbose_name='status')

    class Meta:
        table_name = 'name_model'

coleifer:
Going to pass. Column comments is mysql-specific, and the varchar max length is negligible in my view, since adding it to the introspected field doesn't change any behavior (and it usually isn't enforced anywhere).

PR #2708, opened 2023-04-14
update pwiz add fields attr

GuoGuo54:
Issues #2707 :  https://github.com/coleifer/peewee/issues


coleifer:
Going to pass. `column_comment` isn't supported by postgres, btw.

PR #2710, opened 2023-04-20
feature: When `.in_(None)` is executed, the null condition should be …

ponponon:
When `.in_(None)` is executed, the null condition should be ignored as `0 = 1`


coleifer:
No, I think `.in_(None)` is incorrect since `None` is not an iterable - going to pass on this.

PR #2738, opened 2023-06-01
feat: customize query.dicts() column keys

gshmu:
customize dicts() key example

```python
from peewee import AutoField, ForeignKeyField, Model


class File(Model):
    id = AutoField()


class Page(Model):
    id = AutoField()

    file = ForeignKeyField(File, backref="pages")


if __name__ == "__main__":
    print(list(Page.select(Page.file).dicts().execute())[0])
    print(list(Page.select(Page.file_id).dicts().execute())[0])
    print(list(Page.select(Page.file_id, Page.file).dicts().execute())[0])
```
the origin Page.file is Page.file_id, tag the Page.file_id for the dicts() results.

output:
```
{'file': 1}
{'file_id': 1}
{'file_id': 1, 'file': 1}
```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -5457,7 +5457,7 @@ class ForeignKeyField(Field):
             self.backref = '%s_set' % model._meta.name
 
         if set_attribute:
-            setattr(model, self.object_id_name, ObjectIdAccessor(self))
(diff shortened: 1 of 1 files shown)


coleifer:
Going to pass. You can always specify an explicit alias if you want to control these.

PR #2742, opened 2023-06-06
Prefetch

gshmu:
fix: prefetch peewee.ProgrammingError


coleifer:
Can you share a bit more info, such as how the programming error comes about, and what specifically this is intended to fix? Perhaps you could add a testcase that shows the failure that occurs before this fix is applied, but if you can provide me with enough info I could add it myself (assuming these changes are correct).

coleifer:
Merged (tentatively).

Issue #1644, opened 2018-06-26
reflection got wrong mysql blob field type


GrayMissing:
using pwiz to reflect created mysql table, I found blob column got a TextField defintion, which can lead to codec error. I don't think it's a proper way to decode byte_array using utf-8.

I read the issue before #1034 , but django's TextField seems more like BlobField in peewee...

coleifer:
I'm confused by your issue. What exactly is the problem you are reporting? And if you read #1034, what are you confused about?

GrayMissing:
pwiz reflect mysql blob column to TextField, and TextField decode byte value using utf-8

coleifer:
Yeah, as it states in #1034, Peewee cannot differentiate blob from text due to a limitation in the introspection provided by pymysql. So you need to go through and change the TextFields to BlobFields.

Issue #1034, opened 2016-08-10
pwiz not output models.py  with Mysql BLOB


rfyiamcool:
pwiz not output models.py  with Mysql BLOB? 

parse error, My mysql filed is blob,  pwiz output TEXTFIELD .

Thanks U.

coleifer:
The mapping in the `playhouse.reflection` module explicitly associates MySQL BLOB columns with `TextField`s. I wish I had left a comment as to why it is that way... There seems to be some odd behavior with the way MySQL / Python handle BLOBs compared to the other database engines Peewee supports out-of-the-box (the blob field test cases are skipped for MySQL). I've made a couple changes and will let the continuous integration run the tests and see what shakes out from there.

coleifer:
If you take a look here, it seems to be a limitation of the Python MySQL drivers:

https://github.com/django/django/blob/master/django/db/backends/mysql/introspection.py#L15-L37

Note how all the `BLOB` types get mapped to `TextField` -- this is also the behavior of pymysql:

https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/constants/FIELD_TYPE.py

So rather than assuming BlobField everywhere, peewee does what Django does and assumes TextField. You can go back through the code and make replacements where needed, of course.

PR #2752, opened 2023-07-04
MySQL blob field incorrectly converted to TextField

viorels:
Not sure if this is enough, but it works for me. Should there be a differentiator between the different blob types?


coleifer:
See:

* https://github.com/coleifer/peewee/issues/1034#issuecomment-239906596
* https://github.com/coleifer/peewee/issues/1644

PR #2755, opened 2023-07-12
Add SKIP LOCKED support for SELECT FOR UPDATE

mark99i:
The SKIP LOCKED clause in SELECT FOR UPDATE statement is used to improve concurrency in concurrent transactions by allowing a transaction to skip over rows that are already locked by other transactions. This helps to reduce contention and potential delays when multiple transactions are attempting to update the same set of rows simultaneously. By skipping locked rows, transactions can proceed independently and avoid unnecessary waiting, leading to improved performance and throughput.

https://mariadb.com/kb/en/select/#skip-locked
https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html#innodb-locking-reads-nowait-skip-locked

I implemented parameter support the same way you implemented NOWAIT support

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -1781,17 +1781,22 @@ class WindowAlias(Node):
 
 
 class ForUpdate(Node):
-    def __init__(self, expr, of=None, nowait=None):
+    def __init__(self, expr, of=None, nowait=None, skip_locked=False):
         expr = 'FOR UPDATE' if expr is True else expr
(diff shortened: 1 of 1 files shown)


coleifer:
This isn't really necessary as you can pass whatever you want into the `.for_update()` method, e.g.:

```
Reg.select().for_update('for update skip locked').sql()
# Becomes:
SELECT ... FROM `reg` for update skip locked
```

PR #2756, opened 2023-07-12
Add JSONField with JSON_EXTRACT

mark99i:
When working with JSON data, you need a way to fetch them
I have written a `jextract `function that uses SQL `JSON_EXTRACT` for this field so that it can be easily searched by it:

```
class TestModel(Model):
    id: int | AutoField = AutoField()
    data: dict | JSONField = JSONField()

    class Meta:
        database = dbhandle

obj = TestModel.get_or_none(TestModel.data.jextract('$.my_obj.nested_variable') == 10)
```

I also added the possibility of using json indent (formatting) and escaping non-ascii characters.
```
dbhandle.json_ensure_ascii = True # default
dbhandle.json_use_detailed = False # default
```

### DIFF
--- a/playhouse/fields.py
+++ b/playhouse/fields.py
@@ -1,3 +1,5 @@
+import json
+
 try:
     import bz2
 except ImportError:
@@ -11,7 +13,7 @@ try:
 except ImportError:
     import pickle
 
-from peewee import BlobField
+from peewee import BlobField, TextField, Function
 from peewee import buffer_type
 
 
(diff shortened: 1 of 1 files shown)


mark99i:
I'm sorry, I noticed [this](https://github.com/coleifer/peewee/commit/a4b2a7e3a7769f5ede3fefec3a50022f802e71b3) commit late. Perhaps it is worth considering my experience to improve the class in mysql_ext.py

coleifer:
You are welcome to subclass the class you linked in your comment and add your own helper methods for `extract`, etc. Will pass on merging this change.

PR #2769, opened 2023-08-18
Cover transaction begin in ReconnectMixin

nomme:
**The Problem:**
`playhouse.shortcuts.ReconnectMixin` only supports reconnecting on `execute_sql` calls. For applications that always start an event loop by opening a transaction, this is of no use as `begin` uses `execute` (unless you're SQLite).

**The Solution:**
Given that no transaction may have started when calling `begin` it's worth checking if the connection should be reestablished before surfacing the exception. For nested transactions, the current `in_transaction` check remains to not silently reconnect.

There is the case of SQLite that calls `execute_sql` from `begin` so the `_reconnect` function would be called twice. Not ideal, but also not a big deal.

Feedback welcome.


coleifer:
This looks good, thank you!

PR #2771, opened 2023-08-22
update multi types

gshmu:
dict is super of set, its multi types

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -166,7 +166,7 @@ if sys.version_info[0] == 2:
     buffer_type = buffer
     izip_longest = itertools.izip_longest
     callable_ = callable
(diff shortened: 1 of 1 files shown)


coleifer:
What deficiency/bug does this fix?

coleifer:
I believe this patch is incorrect -- I'm not sure what you were attempting to do, so it's unclear to me why you added this -- but the `multi_types` are used for iterables passed as values. For dicts this would just iterate the keys, which seems unlikely to be correct from an end-user perspective.

gshmu:
`Model.field.in_(dict)` instead of `Model.field.in_(set(dict))` of `Model.field.in_(list(dict.keys()))`.

iterate dict ust iterate the keys in python is OK.

PR #2801, opened 2023-10-26
Fix SyntaxWarning: invalid escape sequence

vkrizan:
Fixes majority of
```
SyntaxWarning: invalid escape sequence '\('
```

PS: There is one false positive at https://github.com/coleifer/peewee/blob/8e539def877b6f3efe094e7e15434795bcbc7f19/tests/__init__.py#L79


coleifer:
Thank you!

PR #2802, opened 2023-10-31
implement operator methods of BigBitFieldData to make it easier to use

nyaoouo:


coleifer:
@nyaoouo - I merged a similar patch. The `__i...___` methods as you implemented them did not seem to be working correctly, by the way, and your patch did not include a single testcase. You can check the differences between your code and what I committed here: 83de3b67a20598deb20b3fd3d08c8325866b33fc -- in the future, please submit tests if you want to get a change merged.

PR #2803, opened 2023-11-02
add database table column default value

b40yd:
In special scenarios, data is not written through Peewee, and there are compatibility issues with writing data.

for example:
> Mainly rolled back after software version upgrade, without deleting new fields, 
> without default values can result in data write failure.

E.g:
```python
class DemoModel(Model):
    name = CharField()
    new_demo_test = IntegerField(default=0)
    
    class Meta:
        database = database
        table_name = 'demo'
```


coleifer:
This patch was missing tests, docs and compatibility w/MySQL and Sqlite. See 7e2c227

PR #2804, opened 2023-11-03
Update docs.

b40yd:


coleifer:
Was missing docs/explanation for the `default=` parameter. Merged an equivalent patch.

PR #2810, opened 2023-12-01
Fix non-process bytes when Handling Inputs of Unequal Length

nyaoouo:
If self._buffer and data have different lengths, the shorter one gets truncated due to the use of zip in the loop, meaning parts of the longer input won't be processed.


coleifer:
We can just use `zip_longest`. Also added tests.

Issue #2838, opened 2024-02-16
[question] How to filter against boolean fields?


blag:
Various Python linters try to force comparisons with booleans to use the `is` operator, as in: `variable is True` or `variable is False`. But the Peewee documentation on [`query operators`](https://docs.peewee-orm.com/en/latest/peewee/query_operators.html#query-operators) doesn't contain an `is` operator, and a given example even references the `==` operator:

```python
# Find any Tweets by users who are not admins (NOT IN).
admins = User.select().where(User.is_admin == True)
```

That seems to imply that we should use the equality operator, even for boolean fields, but it isn't made explicit (as far as I have found).

Should Peewee users always use `==` and just silence warnings/errors from Python linters?

Or is it (also) possible to use the `is` operator? Like this:

```python
# Find any Tweets by users who are not admins (NOT IN).
admins = User.select().where(User.is_admin is True)
```

Or is it possible to just use the boolean fields in a query expression without a comparison operator? Like this:

```python
# Find any Tweets by users who are not admins (NOT IN).
admins = User.select().where(User.is_admin)
```

Apologies if this is already documented and I missed it. If hasn't yet been made explicit, it might be good to throw a note or a footnote near the comparison operators table about this. Happy to contribute a PR once this is cleared up.

Thanks!

coleifer:
Python doesn't support overloading the `is` (identity) test, so you cannot use it in constructing queries.

The following are all valid examples, though:

```python
# Get admins
User.select().where(User.is_admin == True)

User.select().where(User.is_admin)

# Get non-admins
User.select().where(User.is_admin == False)

User.select().where(~User.is_admin)
```

Again we can't use `not`, `or` or `and` because these are not overloaded. So we use the bitwise operators unary negation, bitwise or, bitwise and.

PR #2839, opened 2024-02-16
Better document how to query BooleanFields

blag:
Betterize documentation for querying with `.where()` and for `BooleanField`s.

Ref: #2838


coleifer:
Merged a slightly different patch.

PR #2890, opened 2024-05-10
Fix issue with minute overwriting month in convert_timestamp, python >= 3.12

lanceris:
In `convert_timestamp()` variables for month and minute were the same (`m`), which led to incorrect timestamp parsing.

This PR fixes it by renaming variables to `month` and `minute` respectively, as well as renaming other one-lettered variables in this function to make them more obvious.


coleifer:
Oh man that's awful, thank you!

PR #2898, opened 2024-05-21
Fix operator docs formatting

pylipp:
Otherwise rendered as: Use `== `` and ``!=` for comparing...


coleifer:
Thx, committed an equivalent patch.

PR #2915, opened 2024-07-12
Update db_url.py

chrisribe:
Fixed issue where split would also split passwords containing `?` 
This fix limits the split to only be performed once.


coleifer:
I chose to fix this slightly differently, removing the python 2.6 workaround: https://github.com/coleifer/peewee/commit/73065214abdff535154bb9399f241dc75248fb14

PR #2930, opened 2024-09-16
Add missing python_value for Postgres JSON field.

aaronfultonnz:
I have a need to use table.thaw on a postgres table with a JSON field. The thaw does not work correctly, JSON is inserted as a string and the quotation marks are escaped. I compared postgres to other JSON fields for other database drivers and found there was no python_value method as there was for MySQL and sqlite.

This pull request adds this missing method (taken from MySQL extension)

### DIFF
--- a/playhouse/postgres_ext.py
+++ b/playhouse/postgres_ext.py
@@ -299,9 +299,17 @@ class JSONField(Field):
     field_type = 'JSON'
     _json_datatype = 'json'
 
-    def __init__(self, dumps=None, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


Seluj78:
I think this needs a documentation update as well

coleifer:
I'm marking this as wontfix as it really only seems to occur with thaw + CSV format in the dataset module (when used with a json field). Blindly trying to `json.loads` everything is a no-go from my point-of-view. The other fields use this because they have no "native" json support, but Postgres does via psycopg, so we defer that to the driver.

PR #2943, opened 2024-10-23
Very small copy and paste issue in drop_table doco

timgates42:


coleifer:
Thank you, nice catch.

PR #2961, opened 2024-12-24
 Optimize data migration, the issue of setting default values for new…

b40yd:
Optimize data migration, the issue of setting default values for new fields.

If there are the following data tables in PostgreSQL.

example (refer link:[scanbuf.net](https://www.scanbuf.net/post/postgres/postgres-sql-batch-gen-data/)):
 
```sql
CREATE TABLE public.users (
    id varchar(128) NOT NULL,
    username varchar(64) DEFAULT ''::character varying NOT NULL,
    created_time timestamp NOT NULL,
    updated_time timestamp NOT NULL,
    CONSTRAINT user_pkey PRIMARY KEY (id)
);


CREATE OR REPLACE FUNCTION random_char(chars TEXT) RETURNS CHAR LANGUAGE SQL AS $$
    SELECT substr(chars, floor(random() * length(chars) + 1)::int, 1);
$$;

CREATE OR REPLACE FUNCTION random_string(length INT, chars TEXT) RETURNS TEXT LANGUAGE SQL AS $$
    SELECT string_agg(random_char(chars), '')
    FROM generate_series(1, length);
$$;

INSERT INTO public.users
(id, username, created_time, updated_time)
SELECT
    gen_random_uuid()::character varying,  
    random_string((1 + random() * 20)::int, 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789._+-'),
    NOW() - (random() * (INTERVAL '365 days')),  
    NOW() - (random() * (INTERVAL '365 days')) 
FROM generate_series(1, 10000000)
```

peewee model:

```python
class BaseModel(Model):
    class Meta:
        database = master_db

class User(BaseModel):
    id = CharField(max_length=40, unique=True, primary_key=True)
    username = CharField()
    created_time = DateTimeField()
    updated_time = DateTimeField()
```

Now, I add an new `age` field.

```python
class User(BaseModel):
    id = CharField(max_length=40, unique=True, primary_key=True)
    username = CharField()
    created_time = DateTimeField()
    updated_time = DateTimeField()
    age = IntegerField(default=18)
```

migrate script:

```python
migrator.add_column('user', 'age', pw.IntegerField(default=18))
```

run it. 

it's so so slow...

### DIFF
--- a/playhouse/migrate.py
+++ b/playhouse/migrate.py
@@ -231,8 +231,11 @@ class SchemaMigrator(object):
     def alter_add_column(self, table, column_name, field):
         # Make field null at first.
         ctx = self.make_context()
-        field_null, field.null = field.null, True
 
+        default = field.default
+        if callable_(default):
+            default = default()
+        
         # Set the field's column-name and name, if it is not set or doesn't
         # match the new value.
         if field.column_name != column_name:
@@ -241,9 +244,10 @@ class SchemaMigrator(object):
         (self
          ._alter_table(ctx, table)
          .literal(' ADD COLUMN ')
-         .sql(field.ddl(ctx)))
+         .sql(field.ddl(ctx))
+         .literal(' DEFAULT ')
+         .sql(default))
 
-        field.null = field_null
         if isinstance(field, ForeignKeyField):
             self.add_inline_fk_sql(ctx, field)
         return ctx
@@ -324,13 +328,6 @@ class SchemaMigrator(object):
             raise ValueError('Foreign keys must specify a `field`.')
 
         operations = [self.alter_add_column(table, column_name, field)]
-
-        # In the event the field is *not* nullable, update with the default
-        # value and set not null.
-        if not field.null:
-            operations.extend([
-                self.apply_default(table, column_name, field),
-                self.add_not_null(table, column_name)])
 
         if is_foreign_key and self.explicit_create_foreign_key:
             operations.append(


coleifer:
No plans at the moment to change this. Specifying the default as a constraint is a very different behavior than specifying `default=` in python.

b40yd:
> No plans at the moment to change this. Specifying the default as a constraint is a very different behavior than specifying `default=` in python.

We have also observed a phenomenon where the disk space of a Postgres table significantly expands when there is a large amount of data in the table.

coleifer:
> Specifying the default as a constraint is a very different behavior than specifying default= in python.

Maybe i can improve this by dropping the default after adding the column. On very large tables, specifying a default can be extremely slow, however. I don’t know that there’s a silver bullet here.

b40yd:
> > Specifying the default as a constraint is a very different behavior than specifying default= in python.
> 
> Maybe i can improve this by dropping the default after adding the column. On very large tables, specifying a default can be extremely slow, however. I don’t know that there’s a silver bullet here.

I have rewritten the `add_column` method for peewee migrator in our business code. 

Note: **Only support postgres**.

example:
```python
class PostgreSQLMigrator(PostgresqlMigrator):
    @operation
    def _add_column_and_default(self, table, column_name, field):
        ctx = self.make_context()

        default = field.default
        if callable_(default):
            default = default()

        if field.column_name != column_name:
            field.name = field.column_name = column_name

        (self
         ._alter_table(ctx, table)
         .literal(' ADD COLUMN ')
         .sql(field.ddl(ctx))
         .literal(' DEFAULT ')
         .sql(default))

        if isinstance(field, ForeignKeyField):
            self.add_inline_fk_sql(ctx, field)
        return ctx
    
    @operation
    def add_column(self, table, column_name, field):
        if not field.null and field.default is None:
            raise ValueError('%s is not null but has no default' % column_name)

        is_foreign_key = isinstance(field, ForeignKeyField)
        if is_foreign_key and not field.rel_field:
            raise ValueError('Foreign keys must specify a `field`.')

        operations = [self._add_column_and_default(table, column_name, field)]


        if is_foreign_key and self.explicit_create_foreign_key:
            operations.append(
                self.add_foreign_key_constraint(
                    table,
                    column_name,
                    field.rel_model._meta.table_name,
                    field.rel_field.column_name,
                    field.on_delete,
                    field.on_update))

        if field.index or field.unique:
            using = getattr(field, 'index_type', None)
            operations.append(self.add_index(table, (column_name,),
                                             field.unique, using))

        return operations
```

The expected SQL to be executed is as follows:

```sql
ALTER TABLE demo ADD new_field int DEFAULT 0 NOT NULL;
```

PR #2974, opened 2025-02-20
add unquote_user option

bassettb:
Add the option to unquote the username also, to match the existing unquote_password option.
It is common for a username to contain an @ sign, especially in a major Cloud provider using IAM usernames.

### DIFF
--- a/playhouse/db_url.py
+++ b/playhouse/db_url.py
@@ -39,7 +39,7 @@ def register_database(db_class, *names):
     for name in names:
         schemes[name] = db_class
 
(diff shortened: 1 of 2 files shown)


coleifer:
Nice, thanks!

bassettb:
Any chance of a release with this?

coleifer:
Probably be a bit, I just tagged one.

bassettb:
I don't see a new tag (or release).  It would be awesome if you could create one.

PR #2995, opened 2025-06-01
fix(pwiz): prevent escape sequences in SQLite codegen

senlinjun:
Avoid unintended escape sequences (e.g. \t, \n) when:
- Command-line args contain backslashes
- Database name starts with 't' or 'n' etc.


coleifer:
Going to pass

PR #3002, opened 2025-07-03
increase csv module field_size_limit

### LINKED ISSUE #3001: Increase csv field_size_limit to accommodate large fields in the playhouse/DataSet implementation
When thawing using the playhouse/DataSet module, I got the following error:

 __csv.Error: field larger than field limit (131072)_

I do need to store large fields, so wondering if you could consider changing the field_size_limit for the csv module used in dataset.py.

I have tested and it is solved by adding the following to dataset.py on line 28:

`csv.field_size_limit(pow(2, 31) - 1)
`

Thanks

d1ng3s:
would resolve #3001

### DIFF
--- a/playhouse/dataset.py
+++ b/playhouse/dataset.py
@@ -25,7 +25,7 @@ else:
     def open_file(f, mode, encoding='utf8'):
         return open(f, mode)
 
-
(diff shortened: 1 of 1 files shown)


coleifer:
I'm going to pass - you can apply this in your own code.

PR #3042, opened 2026-04-20
Add documentation for scalars method

JacksonKontny:
@coleifer I'm happy to create an issue if you prefer but I figured this is a pretty small MR with no functional changes that simply closes a gap in the documentation.  LMK what you think.


coleifer:
Thank you!

PR #3056, opened 2026-06-30
Escape double quotes in table names in SQLite introspection

mahirhir:
The SQLite introspection methods interpolate the table name straight into the `PRAGMA` statement without escaping an embedded double quote:

```python
cursor = self.execute_sql('PRAGMA "%s".table_info("%s")' % (schema, table))
```

c2cf25157 added double-quote escaping for the `schema` name in these same methods, and `MySQLDatabase.get_indexes`/`get_primary_keys` already escape the interpolated table name before `SHOW INDEX FROM`. The table name in the SQLite methods was the one identifier left unescaped, so a name containing a `"` breaks out of the quoted identifier:

```
PRAGMA "main".table_info("a"b")   ->  OperationalError: near "b": syntax error
```

This affects `get_indexes`, `get_columns`, `get_primary_keys` and `get_foreign_keys`. The fix escapes the name where it is interpolated, matching the existing `schema` escaping. It is applied at the interpolation site rather than by reassigning `table`, because these methods also pass the raw name as a bound parameter (`get_indexes`) and return it in the metadata (`get_columns`, `get_foreign_keys`).

Added a SQLite regression test that introspects a table whose name contains a `"`. It fails on master with the syntax error shown above and passes with the change. `tests/db_tests.py` introspection, reflection and migration tests stay green.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -4259,7 +4259,7 @@ class SqliteDatabase(Database):
         # Determine which indexes have a unique constraint.
         unique_indexes = set()
         cursor = self.execute_sql('PRAGMA "%s".index_list("%s")' %
-                                  (schema, table))
+                                  (schema, table.replace('"', '""')))
         for row in cursor.fetchall():
             name = row[1]
             is_unique = int(row[2]) == 1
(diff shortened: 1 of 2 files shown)


coleifer:
Thanks bro, yeah I fixed this slightly differently by introducing a helper and also nabbing one spot you missed w/r/t the index name.

Fixed in: a5a7c2a09099943d5bd10a768fb88bc1cead626a

PR #3059, opened 2026-07-12
Fix missing OverflowError in IntegerField and FloatField adapt()

koteshyelamati:
`int(float('inf'))` raises `OverflowError`, not `ValueError` or `TypeError`. Similarly, `float(10**10000)` raises `OverflowError: int too large to convert to float`.

Both `IntegerField.adapt()` and `FloatField.adapt()` currently catch only `(ValueError, TypeError)`, so passing extreme values propagates an unhandled `OverflowError` instead of returning the original value gracefully.

**Fix:** Add `OverflowError` to the except tuple in both methods.

Reproduction:
```python
from peewee import IntegerField, FloatField
IntegerField().adapt(float('inf'))  # OverflowError: cannot convert float infinity to integer
FloatField().adapt(10**10000)       # OverflowError: int too large to convert to float
```

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -5564,7 +5564,7 @@ class IntegerField(Field):
     def adapt(self, value):
         try:
             return int(value)
-        except (ValueError, TypeError):
+        except (ValueError, TypeError, OverflowError):
             return value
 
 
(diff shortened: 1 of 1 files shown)


coleifer:
The Overflow is correct in this case, and working as intended.

koteshyelamati:
Thank you for the clarification, @coleifer! I understand — the `OverflowError` is intentional behavior here and shouldn't be silently caught. I'll close this PR.

coleifer:
Bro are you using ai for the responses too?

PR #3063, opened 2026-07-27
Add 'null' field to _FieldKwargs type

pylipp:
In the following example containing self-referential FKs, mypy complains about finding no overload variant for ForeignKeyField:
```python
from peewee import ForeignKeyField, Model

class Human(Model):
    parent = ForeignKeyField("self", null=False)
```
With the proposed update, the warning disappears.
It seems like `null` was missed in the list of inputs for the `Field.__init__` method although the input is defined in the [actual implementation code](https://docs.peewee-orm.com/en/latest/peewee/api.html#Field).


coleifer:
Nah the omission is intentional we do all that via overloads. I’ll get a patch together for this particular case.

coleifer:
Good to go w/this change: f68e857983dc19b5b32d767aa6e7de4c84953b19

PR #3070, opened 2026-08-14
Fix JSONField when the database proxy is initialized before the model

NotAFlightRisk:
Hit this with the proxy initialised before the models module gets imported. Every `JSONField` on those models quietly stopped working.

```python
proxy = DatabaseProxy()
proxy.initialize(SqliteDatabase(':memory:'))  # swap these two round and it's fine

class Note(Model):
    data = JSONField()
    class Meta:
        database = proxy

Note.create(data={'k': 1})
# peewee.ProgrammingError: Error binding parameter 1: type 'dict' is not supported
```

`Note.data['k']` gives `AttributeError: 'NoneType' object has no attribute 'extract'`, and scalars are worse - `data='hello'` stores the bare text with no error at all, so `json_valid()` on the column comes back 0.

`attach_callback()` only appends, so a proxy thats already initialised never fires the hook, and `bind()` was passing `None` either way. `Metadata.set_database()` already only nulls a proxy when `obj is None`, so I've made `bind()` do the same.

Only bites since JSONField moved onto that hook - BlobField's fallback was `bytearray`, which every driver takes.

Tested on sqlite only (`python runtests.py`, 1826 pass). The postgres and mysql helpers go through the same hook so I'd expect those fine, but I've not run them.

### DIFF
--- a/peewee.py
+++ b/peewee.py
@@ -5872,14 +5872,13 @@ class FieldDatabaseHook(object):
         raise NotImplementedError('Subclasses must implement')
 
     def bind(self, model, name, set_attribute=True):
-        if model._meta.database is not None:
-            if isinstance(model._meta.database, Proxy):
-                model._meta.database.attach_callback(self._db_hook)
-                self._db_hook(None)
-            else:
-                self._db_hook(model._meta.database)
-        else:
-            self._db_hook(None)
+        database = model._meta.database
+        if isinstance(database, Proxy):
+            database.attach_callback(self._db_hook)
+            # An uninitialized proxy is treated as None.
+            if database.obj is None:
+                database = None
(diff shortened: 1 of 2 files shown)


coleifer:
Oh LLM prose…what was life before these offenses.

coleifer:
Fixed by 408aa4c

NotAFlightRisk:
tung-sahur-side-eye-tung-tung-tung-sahur

Thanks for the fixx anyway :)

Yeah, i'm just a smooth brained low-life, incapable of doing anything without chatgpt nowadays, so your response is fair 👍

coleifer:
welcome, you're among friends



patch looked good, btw, just wanted to refactor the test up a little.

PR #3072, opened 2026-08-15
Add SmallAutoField

johanjk:
Been trying to save a questionable amount of bytes using `smallserial`.
Now you can too!


coleifer:
I honestly don’t think it’s worth it. I think you’d only see savings if you had a ton of tables since a single table is limited to how much space it can even save. Gonna pass. In 16 years nobody has needed it before.

johanjk:
Understandable. Not many merits outside of the human irk of having symmetry and completeness in the framework.

PR #3082, opened 2026-09-19
Avoid collisions in generated result-column names

### LINKED ISSUE #3081: Generated result-column suffixes can overwrite selected values
On master (`1594d56a90d4674e873324efef561a04aa805331`, Peewee 4.5.1), a generated column suffix can collide with another result column and silently discard a value from dictionary rows.

```python
from peewee import Select, SqliteDatabase, Value

query = Select(columns=[
    Value(1).alias('value'),
    Value(2).alias('value'),
    Value(3).alias('value_2'),
]).bind(db)
print(query.dicts().get())
# {'value': 1, 'value_2': 3}
```

All three values should be available under distinct keys. `dedupe_columns()` tracks the original names but does not reserve generated names. The same issue occurs when `value_2` precedes the second `value`.

Reproduced with Python 3.13.15 and SQLite 3.53.1. A regression covering model and plain select queries, dictionaries, objects and named tuples passes after checking each generated name against names already used.

jakezwang:
Fixes #3081.

Keep generated column names in the duplicate-name map and advance the suffix until the name is unused. Queries with aliases such as `value, value, value_2` then retain every value in dictionary and object rows.

Cover both alias orders and nested suffix collisions for model/plain queries, dictionaries, objects and named tuples. Document duplicate-key naming.

Validation: 47 result tests; full SQLite suites on Python 3.9 and 3.11–3.14; documentation and stub/type checks passed. Python 3.8 has one FTS query failure also reproduced on the unchanged base.


coleifer:
Bro do you even use peewee?

PR #3084, opened 2026-09-19
Preserve lazy loading for unselected outer joins

### LINKED ISSUE #3083: Unselected LEFT OUTER JOIN suppresses foreign-key lazy loading
With Peewee 4.5.1, adding a LEFT OUTER JOIN without selecting columns from the
joined model makes a valid foreign-key relationship return `None`. The foreign-key
ID is still present and the related row exists. This also affects model aliases.

Reproduced on upstream master at `1594d56a90d4674e873324efef561a04aa805331`
with Python 3.14.2 and SQLite 3.50.4.

```python
from peewee import JOIN, ForeignKeyField, Model, SqliteDatabase, TextField

class Base(Model):
    class Meta:
        database = db

class Course(Base):
    name = TextField()

class Invoice(Base):
    course = ForeignKeyField(Course, null=True)

db.create_tables([Course, Invoice])
course = Course.create(name='Existing course')
Invoice.create(course=course)
invoice = Invoice.select().join(Course, JOIN.LEFT_OUTER).get()
print(invoice.course_id)  # 1
print(invoice.course)     # Actual: None; expected: related Course instance.
assert invoice.course is not None
```

The last line raises `AssertionError`. Without the join, the relationship loads
normally. Selecting both `Invoice` and `Course` also works.

`ModelCursorWrapper.process_row()` treats an unselected join destination as a
verified missing row and caches `None` in `invoice.__rel__['course']`. That prevents
normal lazy loading. The proposed fix records the selected sources and skips
relationship assignment when neither the destination nor any joined descendant
is selected. Selected NULL results and selected descendants keep their existing
behavior.

Seven regression tests cover ordinary and aliased lazy loading, eager loading,
NULL foreign keys, restrictive unmatched joins preserving FK IDs, and present or
NULL selected descendants. Before the fix, the ordinary and aliased lazy-loading
tests fail; after the fix all seven pass.

Related: #3060 discusses selected outer-join results. This report concerns a joined source with no selected columns; the proposed fix preserves the selected-result behavior discussed there.

igoncharov85:
A LEFT OUTER JOIN that selects no columns from the related model currently suppresses foreign-key lazy loading. Accessing the relationship returns `None`, even when the foreign-key ID is present and the related row exists. This occurs in Peewee 4.5.1 and current master, including when the joined model is aliased.

For example, given an `Invoice` model with a `course = ForeignKeyField(Course)` field pointing to an existing course:

```python
invoice = Invoice.select().join(Course, JOIN.LEFT_OUTER).get()
invoice.course_id  # The existing course's ID.
invoice.course     # Currently None; should lazy-load the Course instance.
```

The query selects only invoice columns. During row hydration, Peewee nevertheless treats the unselected course as a missing joined row and caches `None` for the relationship, preventing lazy loading.

This change leaves the relationship unset when neither the joined source nor any of its joined descendants contributes selected columns. Normal foreign-key lazy loading can then resolve it. Selected related rows, selected NULL results, and joins with selected descendants retain their existing behavior. SQL generation and public APIs are unchanged.

Includes seven regression tests in `tests/results.py`, with query-count assertions for lazy and eager loading, plus relationship documentation and a changelog entry. The tests also cover aliases, NULL foreign keys, restrictive unmatched joins preserving foreign-key IDs, and selected descendants.

Validation on Python 3.14.2 / SQLite 3.50.4:

- Before the fix: seven regression tests run, two expected failures.
- `python3 runtests.py -e sqlite models results prefetch_tests with_related_tests`: 485 tests run, 63 skipped, no failures.
- `python3 runtests.py -e sqlite`: 1,832 tests run, 180 skipped, no failures.
- `git diff --check`: clean.

Optional extension dependencies were unavailable in this environment; PostgreSQL, MySQL, and other backend configurations were not run.

Fixes #3083.


coleifer:
My friend, do you use peewee?

igoncharov85:
Yes, I use peewee.

This issue appeared in 4.2.2:
* 4.2.0–4.2.1: pass.
* 4.2.2–4.2.5: fail; the join incorrectly clears the foreign-key ID.
* 4.2.6–4.5.1: fail; the ID is preserved, but the relationship is cached as None, preventing lazy loading.

coleifer:
Doubt.

igoncharov85:
I've been using peewee for the last 3 years, and my project is covered with a lot of unit tests.
It used to work without this issue till 4.2.2
I still keep 4.2.1 due to this regression bug.

coleifer:
In that case i will get into patching this. I apologize for the skepticism, I’ve been getting a steady stream of ai prs, most of which are junk. This one felt valid from the get go but I wasn’t clear if it was theoretical or actually breaking someone’s code.

coleifer:
This is fixed in fa29ff5

igoncharov85:
Thank you very much! Looking forward for the next Peewee release.
Thx

